PR #1, opened 2016-05-13
Enable use of the libv8 gem

ignisf:
libv8 5.0.71.48.1beta2 is up for testing, got mini_racer to compile and run tests with the changes below.

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -1,8 +1,8 @@
 #include <stdio.h>
 #include <ruby.h>
 #include <ruby/thread.h>
-#include <v8.h>
-#include <libplatform/libplatform.h>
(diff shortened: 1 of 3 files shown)


ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -1,8 +1,8 @@
 #include <stdio.h>
 #include <ruby.h>
 #include <ruby/thread.h>
-#include <v8.h>
-#include <libplatform/libplatform.h>
+#include <include/v8.h>
+#include <include/libplatform/libplatform.h>
```
ignisf: not sure why is "include/" necessary here, maybe @cowboyd could chime with an opinion if it's OK to change the include paths in libv8

SamSaffron:
wow awesome!!!! Thanks heaps

PR #12, opened 2016-05-25
Add js Date marshalling from rb Time 

seanmakesgames:


seanmakesgames:
also, this is my first open-source contribution-- feedback is much appreciated!

SamSaffron:
this looks great! it would be nice to support DateTime -> JS as well if possible cause it is fairly common for Rails apps, that said perfectly happy to merge this and great change!!! 

👍

seanmakesgames:
Thanks @SamSaffron I can add that probably as well. Discussion needs to be had on that however, because it would be only one-way atm (there's only Date() on the JS side, so we'd be coming back as Time)

SamSaffron:
I think one way is fine for this case, the timezone thing gets a bit hairy
so whatever we do we should document in the readme.

PR #14, opened 2016-05-26
Add 64bit int support

seanmakesgames:
converts to a double to get it across the line and back. (ala therubyracer)
trying to do this without the patch will throw a can't convert int64 to int32 exception

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -221,17 +221,18 @@ static Handle<Value> convert_ruby_to_v8(Isolate* isolate, VALUE value) {
     VALUE pair;
     int i;
     long length;
+    long fixnum;
     VALUE klass;
 
     switch (TYPE(value)) {
     case T_FIXNUM:
-        length = NUM2INT(rb_funcall(value, rb_intern("size"), 0));
-        if (length > 4)
+        fixnum = NUM2LONG(value);
+        if (fixnum > INT_MAX)
         {
-            return scope.Escape(Number::New(isolate, NUM2DBL(value)));
+            return scope.Escape(Number::New(isolate, (double)fixnum));
         }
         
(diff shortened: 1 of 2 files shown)


ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -225,7 +225,13 @@ static Handle<Value> convert_ruby_to_v8(Isolate* isolate, VALUE value) {
 
     switch (TYPE(value)) {
     case T_FIXNUM:
-	return scope.Escape(Integer::New(isolate, NUM2INT(value)));
+        length = NUM2INT(rb_funcall(value, rb_intern("size"), 0));
+        if (length > 4)
+        {
+            return scope.Escape(Number::New(isolate, NUM2DBL(value)));
```
SamSaffron: I wonder, why not simply always give JavaScript Number::New unconditionally, that way we avoid the length check?

seanmakesgames: We could do that, but then we are marshalling in as double-precision floating point all the time, which could cause other issues. I wonder why v8 provides an "Integer" class but not an int64 class

seanmakesgames: Also, I'm not sure where the code came from initially, and why it was not done that way to begin with? Also NUM2LONG exists, but doesn't work with the Integer class which is looking for an int32

SamSaffron: I see, so we lose some precision here... ok, my call is to use NUM2LONG, then we avoid the funccall, simply check if its larger than int max value if larger then convert to double etc...

seanmakesgames: I'm on board with that. I'll update the branch

seanmakesgames:
@SamSaffron rdy for review.

SamSaffron:
this looks great! thanks

PR #16, opened 2016-05-26
DateTime marshalling

seanmakesgames:


ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -620,6 +627,8 @@ extern "C" {
 	rb_define_private_method(rb_cContext, "eval_unsafe",(VALUE(*)(...))&rb_context_eval_unsafe, 1);
 	rb_define_private_method(rb_cExternalFunction, "notify_v8", (VALUE(*)(...))&rb_external_function_notify_v8, 0);
 	rb_define_alloc_func(rb_cExternalFunction, allocate_external_function);
+    
+    rb_cDateTime = rb_const_get(rb_cObject, rb_intern("DateTime"));
```
SamSaffron: hmmm what happens if DateTime is not defined? 

```
irb(main):001:0> DateTime
NameError: uninitialized constant DateTime
```

seanmakesgames: I'm not sure, we're not calling any methods on it, just the const_get of it initially. I'll check on it

seanmakesgames: Adding in this line before creating the context in code doesn't seem to affect the tests at all, I must not be doing it correctly.
`
    Object.send("remove_const", :DateTime)
`
What cases might this even happen? DateTime is part of the standard lib of ruby right? Would someone have to do this on purpose?

SamSaffron:
Yeah I see, you must `require "date"` to load up DateTime. so this is bit of a pickle, if for some reason it is not required when we require mini_racer we are in trouble. so we have to require 'date' before looking up that constant (from c, cause in ruby it is too late)

seanmakesgames:
Do you know how we can write the test so that tests the no-datetime scenario?

seanmakesgames:
Spent some time, doesn't look like you can write a test for this scenario: 
Init for the extension happens before the test code runs at all. (which I believe is a good thing) :P
I'm testing it manually in the c-code and putting together a fix.

seanmakesgames:
Update: Looks like this is what happens when you don't have datetime and mini_racer gets 'require'd 
```
/Users//mini_racer/lib/mini_racer.rb:2:in `require': uninitialized constant DateTime (NameError)
	from /Users//mini_racer/lib/mini_racer.rb:2:in `<top (required)>'
	from /Users//mini_racer/test/test_helper.rb:3:in `require'
	from /Users//mini_racer/test/test_helper.rb:3:in `<top (required)>'
	from /Users//mini_racer/test/mini_racer_test.rb:1:in `require'
	from /Users//mini_racer/test/mini_racer_test.rb:1:in `<top (required)>'
	from /Users//.rvm/gems/ruby-2.2.4/gems/rake-10.5.0/lib/rake/rake_test_loader.rb:15:in `require'
	from /Users//.rvm/gems/ruby-2.2.4/gems/rake-10.5.0/lib/rake/rake_test_loader.rb:15:in `block in <main>'
	from /Users//.rvm/gems/ruby-2.2.4/gems/rake-10.5.0/lib/rake/rake_test_loader.rb:4:in `select'
	from /Users//.rvm/gems/ruby-2.2.4/gems/rake-10.5.0/lib/rake/rake_test_loader.rb:4:in `<main>'
```

seanmakesgames:
Ok. updated the branch. You can test by adding 
`Object.send("remove_const", :DateTime)`
into test_helper, before the mini_racer require

SamSaffron:
I think this is ok, not super ideal but not sure anything is. 

If mini_racer is included prior to datetime being around rb_cDateTime is set to something undefined, if it later is required and exists it no longer auto-corrects. 

I think a reasonable compromise (so you don't have to re-check each time you convert stuff) would be to do the lookup on object initialization (if it is set to say NULL) and default have it set to NULL

that way there is an infinitely small additional cost when you create a context, but ordering of stuff in the Gemfile will not matter.

seanmakesgames:
Which function block is "object initialization"? I can definitely move the code there, but I assume you have a specific place in mind. :)

SamSaffron:
the "allocate" function is probably simplest spot to add it, just after it
creates the isolate etc.

seanmakesgames:
I think I got it. With new location, I was able to write a test against it as well.

ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -589,6 +596,11 @@ VALUE allocate(VALUE klass) {
     context_info->context = new Persistent<Context>();
     context_info->context->Reset(context_info->isolate, context);
 
+    if (rb_funcall(rb_cObject, rb_intern("const_defined?"), 1, rb_str_new2("DateTime")) == Qtrue)
```
SamSaffron: minor change ... check if it is already set, set rb_cDateTime to NULL by default, only do the funcall if its NULL

SamSaffron:
sorry for the huge delay here, been struggling to find time this week! 

except for the minor comment this looks fine!

seanmakesgames:
It's no problem @SamSaffron -- I'm mostly set up on my end (this was a feature you requested) :D
I do have some new questions though which I'll post to issues.

SamSaffron:
this looks good! thanks!

PR #19, opened 2016-06-10
Adjust header inclusions for changes in V8 5.1.11

ignisf:
Since [1] V8 finally assumes that v8/include is in the include path.

[1] https://chromium.googlesource.com/v8/v8/+/e8c914f18fd88cd58aaaf5d8cee53b9e57477b52

**merge after releasing a version with #20 merged**

### DIFF
--- a/mini_racer.gemspec
+++ b/mini_racer.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "minitest", "~> 5.0"
   spec.add_development_dependency "rake-compiler"
 
-  spec.add_dependency 'libv8', '~> 5.0', '< 5.1.11'
+  spec.add_dependency 'libv8', '~> 5.1'
   spec.require_paths = ["lib", "ext"]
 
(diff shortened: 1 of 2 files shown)


mini_racer.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "minitest", "~> 5.0"
   spec.add_development_dependency "rake-compiler"
 
-  spec.add_dependency 'libv8', '~> 5.0'
+  spec.add_dependency 'libv8', '~> 5.1.281.59.0beta3'
```
ignisf: The version constraint should probably be more strict. How much more strict - I have no idea.

SamSaffron: I think we can err on being a bit more practical here, just use ~>5.1 ... there is no release of 5.1 that will not cover this fix

SamSaffron:
ok 0.1.4 is released, this needs a rebase and we should be good :+1:  thanks heaps

ignisf:
:shipit:

SamSaffron:
awesome thanks @ignisf !

PR #20, opened 2016-06-14
Set an upper limit on the version of libv8

ignisf:
OK, so here's the issue.

In the current state of the gems, if we don't synchronize the releases of a new version of mini_racer with https://github.com/discourse/mini_racer/pull/19 and the new version of libv8, new mini_racer installs will fail.

I suggest:

1. You release a new version of mini_racer that sets this upper limit on the libv8 version
2. I release the new libv8 version
3. You release a new version of mini_racer with https://github.com/discourse/mini_racer/pull/19 merged.


SamSaffron:
cool, doing a new gem release now

PR #22, opened 2016-06-21
Making it possible to share isolates between contexts

wk8:
Allocating and then freeing V8 `Isolate`s turns out to be pretty expensive at times.

This patch makes sharing isolates possible by introducing a new `MiniRacer::Isolate` object,
which can be initialized with a `MiniRacer::Snapshot` object.
That isolate object can then be used to create a new context: `MiniRacer::Context.new(isolate: isolate)`
Of course that must be used with special care (probably not used at all, actually)
in multi-threaded code, but works with no particular caveats otherwise.

Example snippet:
```
snapshot = MiniRacer::Snapshot.new(my_snippet)
isolate = MiniRacer::Isolate.new(snapshot)

# now that isolate can be shared across as many contexts as desired
context1 = MiniRacer::Context.new(isolate: isolate)
context2 = MiniRacer::Context.new(isolate: isolate)
```

This patch introduces no breaking changes; in particular, contexts still get their
own isolate by default.

Particular care has been taken for memory management, with multiple tests on that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,23 @@ are fixed once the snapshot has been captured. They are no longer really random
 nor reflect the current time.
 ```
 
+Also note that snapshots can be warmed up, using the `warmup!` method, which allows you to call functions which are otherwise lazily compiled to get them to compile right away; any side effect of your warm up code being then dismissed. [More details on warming up here](https://github.com/electron/electron/issues/169#issuecomment-76783481), and a small example:
+
+```ruby
+
+snapshot = MiniRacer::Snapshot.new('var counter = 0; function hello() { counter++; return "world! "; }')
+
+snapshot.warmup!('hello()')
+
+context = MiniRacer::Context.new(snapshot: snapshot)
+
+context.eval('hello()')
+# => "world! 1"
+context.eval('counter')
+# => 1
+
+```
+
 ### Shared isolates
 
 By default, MiniRacer's contexts each have their own isolate (V8 runtime). For efficiency, it is possible to re-use an isolate across contexts:
@@ -149,6 +166,24 @@ context = MiniRacer::Context.new(isolate: isolate)
 context.eval("hello()")
 # => "world!"
 ```
+
+Re-using the same isolate over and over again means V8's garbage collector will have to run to clean it up every now and then; it's possible to trigger a _blocking_ V8 GC run inside your isolate by running the `idle_notification` method on it, which takes a single argument: the amount of time (in milliseconds) that V8 should use at most for garbage collecting:
+
+```ruby
+isolate = MiniRacer::Isolate.new
+
+context = MiniRacer::Context.new(isolate: isolate)
+
+# do stuff with that context...
+
+# give up to 100ms for V8 garbage collection
+isolate.idle_notification(100)
+
+```
+
+This can come in handy to force V8 GC runs for example in between requests if you use MiniRacer on a web application.
+
+Note that this method maps directly to [`v8::Isolate::IdleNotification`](http://bespin.cz/~ondras/html/classv8_1_1Isolate.html#aea16cbb2e351de9a3ae7be2b7cb48297), and that in particular its return value is the same (true if there is no further garbage to collect, false otherwise) and the same caveats apply, in particular that `there is no guarantee that the [call will return] within the time limit.`
 
 ## Performance
 
--- a/README.md
+++ b/README.md
@@ -88,6 +88,68 @@ puts context.eval("counter")
 
 ```
 
+### Snapshots
+
+Contexts can be created with pre-loaded snapshots:
+
+```ruby
+
+snapshot = MiniRacer::Snapshot.new('function hello() { return "world!"; }')
+
+context = MiniRacer::Context.new(snapshot: snapshot)
+
+context.eval("hello()")
+# => "world!"
+
+```
+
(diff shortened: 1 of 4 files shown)


SamSaffron:
hmmm what advantage does this offer over simply sharing the context? 

Are you able to somehow "reset" an isolate?

wk8:
@SamSaffron : together with snapshots, that allows creating cheap pre-baked contexts really quick, that you can then pollute as much as you want and then discard as a whole when you're done; and then just get a new pre-baked one for the next request (for example).

Even without bringing snapshots into the picture, V8 isolates are not that cheap to create/destroy; in a single-threaded app it makes sense to re-use the same isolate for all your contexts.

As to resetting isolates, not sure what you mean?

SamSaffron:
Sorry I mean does this somehow work?

```
MiniRacer::Context.new(isolate: isolate).eval('a=1')
MiniRacer::Context.new(isolate: isolate).eval('a') => nil
```

wk8:
@SamSaffron : sure does. Isolates don't hold the JS context (that's what contexts do). Isolates are just the runtime framework.

```
irb(main):001:0> require 'mini_racer'
=> true
irb(main):002:0> isolate = MiniRacer::Isolate.new
=> #<MiniRacer::Isolate:0x007fdb6b142618>
irb(main):003:0> MiniRacer::Context.new(isolate: isolate).eval('a=1')
=> 1
irb(main):004:0> MiniRacer::Context.new(isolate: isolate).eval('a')
MiniRacer::RuntimeError: ReferenceError: a is not defined
```

SamSaffron:
Can you test what happens with concurrency? I want us to be safe by default even if it means we need to lock on a mutex that lives on the isolate object

wk8:
@SamSaffron : test what exactly? You shouldn't re-use the same isolate across threads, for sure. But this patch doesn't change that, by default contexts have their own isolate if you don't pass an isolate explicitly when creating them.

As far as testing it still works across threads by default, isn't that what https://github.com/discourse/mini_racer/blob/bddbdcd284f8b1dce8bd0fd303bd15abf23888f9/test/mini_racer_test.rb#L120-L131 does (haven't touched that test in my PR)? Or do you have something else in mind?

SamSaffron:
Something along the lines of this (but with some warmup code)

```
isolate = MiniRacer::Isolate.new
(1..10).times do 
   Thread.new {
        context = MiniRacer::Context.new(isolate: isolate)
        random = SecureRandom.hex
        context.attach("sleep", proc{sleep(0.01)})        
        context.eval("sleep()")
        context.eval("a='#{random}'")
        context.eval("sleep()")
        context.eval("a") => should be same random
   }
end
```

wk8:
@SamSaffron : no, I don't expect that to work :) as mentioned above, the same isolate shouldn't be used concurrently. That's why by default each context still has its own isolate. The expectation is that anyone using the ability to re-use isolate across contexts should know what they're doing.

SamSaffron:
I follow, but why have that restriction? we already protect eval and attach, just instead have the @lock object live on the isolate

https://github.com/discourse/mini_racer/blob/master/lib/mini_racer.rb#L104

wk8:
@SamSaffron : good point, will change

SamSaffron:
also, we got to update the readme with all the new fancy things :) ideally with a practical example or 2 

very happy with this change it unlocks a bunch of possibilities, gives you certainty that code you are running can never interfere with initial state without a performance cost.

wk8:
@SamSaffron : will do too :)

wk8:
Thanks for being so reactive!

wk8:
@SamSaffron : moved the lock to isolates as discussed, see https://github.com/discourse/mini_racer/pull/22/commits/430b080334b78bc6262f8abb223fec1f46f0141e .

Will add README goodies tomorrow.

Other than that, looking good to you?

Thanks!

wk8:
@SamSaffron : README updated, https://github.com/discourse/mini_racer/pull/22/commits/41d4b3df4276e1b3d5fc331f3cf75d264d04ad09

Thanks!

SamSaffron:
give me a few days here, I am taking the rest of the week off, will review
it all on monday, thanks heaps!

wk8:
@SamSaffron : no worries, no rush! Thanks!

SamSaffron:
this is looking really good! 

Can we document warmup and idle_notification in the readme as well?

wk8:
@SamSaffron : sure, will do, most likely tomorrow, busy day here :) thanks!

SamSaffron:
no probs, these are all very exciting changes!

wk8:
@SamSaffron : expanded the README as requested at https://github.com/wk8/mini_racer/blob/3e1c214975bd7aaf5c444291f752f0c819bc35ce/README.md

SamSaffron:
cool, we need a rebase, but once done feel free to merge this in! :+1:

SamSaffron:
Thanks ❤️  ❤️

SamSaffron:
@wk8  do you have a twitter account? wanted to give you a shout out.

wk8:
@SamSaffron : I don't, sorry. But thank you for your great work :)

SamSaffron:
Ouch, stuff appears to be timing out on Mac and failing on Ruby 2, can you
have a look?

wk8:
@SamSaffron : I've seen this failure (https://travis-ci.org/discourse/mini_racer/jobs/141200273) intermittently in dev too, including from master before my changes. As to mac, the job never seems to get started, maybe a travis issue?

ignisf:
https://www.traviscistatus.com/incidents/rqb906yypnk0

wk8:
@SamSaffron : there's a fix for the intermittent failure: https://github.com/discourse/mini_racer/pull/24

PR #23, opened 2016-06-22
Making it possible to set V8 runtime flags

wk8:
Only caveat: such flags can only be set before the V8 platform gets initialized.

Comes in handy to deactivate for example V8's aggressive use of `pthread` which
somehow seems to not play too nice with Unicorn:

```
MiniRacer::Platform.set_flag! '--noconcurrent_recompilation'
MiniRacer::Platform.set_flag! '--noconcurrent_sweeping'
```

Also protecting the platform initialization with a mutex, not sure
how/why current concurrent tests never seem to hit bugs without that mutex.

Added just one test on the new `MiniRacer::Platform.set_flag!` method, and
unsure how to make other tests without making tests run in order (bad),
given that flags can only be set before anything else is done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,47 @@ puts context.eval("counter")
 
 ```
 
+### V8 Runtime flags
+
+It is possible to set V8 Runtime flags:
+
+```ruby
+MiniRacer::Platform.set_flags! :noconcurrent_recompilation, max_inlining_levels: 10
+```
+
+This can come in handy if you want to use MiniRacer with Unicorn, which doesn't seem to alwatys appreciate V8's liberal use of threading:
+```ruby
+MiniRacer::Platform.set_flags! :noconcurrent_recompilation, :noconcurrent_sweeping
+```
+
+Or else to unlock experimental features in V8, for example tail recursion optimization:
+```ruby
+MiniRacer::Platform.set_flags! :harmony
+
+js = <<-JS
+'use strict';
+var f = function f(n){
+  if (n <= 0) {
+    return  'foo';
+  }
+  return f(n - 1);
+}
+
+f(1e6);
+JS
+
+context = MiniRacer::Context.new
+
+context.eval js
+# => "foo"
+```
+The same code without the harmony runtime flag results in a `MiniRacer::RuntimeError: RangeError: Maximum call stack size exceeded` exception.
+Please refer to http://node.green/ as a reference on other harmony features.
+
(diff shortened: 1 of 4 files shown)


SamSaffron:
I really like this, but we got to document it in the readme, ideally with a list of options. 

Also, perhaps this API is a bit cleaner

```
MiniRacer::Platform.set_flags! :harmony, :noconcurrent_sweeping
```

The readme can also demonstrate how :harmony unlocks features

wk8:
@SamSaffron : thanks, will do :)

Also, I kinda like your API, but how do you see it working for flags that take a value? Will change this PR once we've cleared this point

SamSaffron:
perhaps?

```
MiniRacer::Platform.set_flags! :noconcurrent_sweeping, {something: 77}, :harmony
```

wk8:
@SamSaffron : okay, sounds good :) also, what's your take on using the unfortunately "internal" API https://github.com/v8/v8/blob/master/src/flags.h for better testing? (i.e. I could use `static List<const char*>* argv();` and pass innocuous flags such as these to all tests:
```
  --testing_bool_flag (testing_bool_flag)
  --testing_maybe_bool_flag (testing_maybe_bool_flag)
  --testing_int_flag (testing_int_flag)
  --testing_float_flag (float-flag)
  --testing_string_flag (string-flag)
```
And check the flags are set correctly?

The only issue with the plan above being that it might break at each V8 release :/

wk8:
Actually, `--use_strict` and a source that violates it could be used for better tests, better than relying on an internal API

SamSaffron:
I would stay away from internal APIs, feels too risky. Maybe we do advisory
validation in Ruby and print to STDERR if we notice any "unexpected" flags.
That way it will never really break

wk8:
@SamSaffron : V8 actually already prints to stderr if you pass it an unpexected option, I could add that to tests

SamSaffron:
Sure, sounds good.

wk8:
@SamSaffron: added https://github.com/discourse/mini_racer/pull/23/commits/db82900e6cc482541dd425e15cbbd7dbe436d581, but couldn't find a good example for the use of the harmony flag, sorry

SamSaffron:
I can see a few options here:

http://node.green/

Look for the yellow ones in latest with the flag icon, highlighting the ?
shows source

wk8:
@SamSaffron : thanks for the pointer, added a nice example.

SamSaffron:
excellent, also added you write access to mini_racer in case you need any smaller fixes or updates to doco (or I am on holidays :) )

wk8:
Thanks :) !

PR #24, opened 2016-06-29
Fixing a byzantine test failure

wk8:
Whenever `test_datetime_missing` would run before `test_return_date`, tests would fail
because `DateTime` wouldn't be defined any more.


SamSaffron:
Nice one!

PR #36, opened 2016-08-08
Max Memory pre-fatal alloc termination

seanmakesgames:


ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -512,6 +568,11 @@ static VALUE rb_context_eval_unsafe(VALUE self, VALUE str) {
 	if (timeout != Qnil) {
 	    eval_params.timeout = (useconds_t)NUM2LONG(timeout);
 	}
+	eval_params.max_mem = 0;
+	VALUE max_mem = rb_iv_get(self, "@max_mem");
+	if (max_mem != Qnil) {
+	    eval_params.max_mem = (useconds_t)NUM2LONG(max_mem);
```
dylanryan: size_t, not useconds_t

ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -118,6 +120,39 @@ void* breaker(void *d) {
...
+
+    if (action == kAllocationActionAllocate)
+    {
+        callback_eval_params->current_mem += size;
+        
+        printf("\nALLOC %d: %zu", size, callback_eval_params->current_mem);
+    }
+    else if (action == kAllocationActionFree)
+    {
+        callback_eval_params->current_mem -= size;
```
dylanryan: size_t is unsigned, and you are subtracting off some value. If it is ever possible that you track more frees than allocations (i.e. a free of an allocation from just before you started tracking, or similar), current_mem will underflow to $huge_val, and trigger the termination.

seanmakesgames: Doesn't sound like the worst behavior, but I'll probably put an if in there. Reduces performance though. ;)

ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -57,6 +57,8 @@ typedef struct {
     ContextInfo* context_info;
     Local<String>* eval;
     useconds_t timeout;
+    size_t max_mem;
+    size_t current_mem;
```
dylanryan: Not your fault, but this struct is laid out poorly and will almost certainly waste a few bytes on 64-bit machines due to padding. Any 32-bit objects (such as timeout) should come after any 64-bit objects (such as the EvalResult* and these 2 size_t's). Again, not your fault, it was set up that way, just pointing it out ;p

ext/mini_racer_extension/mini_racer_extension.cc
dylanryan: Yeah. Or.... realistically size_t is probably overkill: unless people want to set memory limits of over 2 gigs (!!), using an int (or int32_t to be safe) instead of size_t would probably work. int64_t (either long or long long depending on platform) if you want to pad it out more. Then you can dip into the negatives sanely.

seanmakesgames: Will switch to int64t.

ext/mini_racer_extension/mini_racer_extension.cc
seanmakesgames: will do

dylanryan: I noticed that throughout the codebase, so probably not worth it. Just pointing it out because it is a pet peeve of mine ;p

Basically, the issue is that 64-bit values often must start on a memory address that is a multiple of 64 bits (not true on all platforms, but on some). 32-bit values must always start on a multiple of 32 bits, and so on (generally speaking). 

So, if a struct is laid out as
64 bit
32 bit
64 bit
16 bit
32 bit

the compiler will have to insert padding:

64 bit
32 bit
[waste 32 bits to align the next item]
64 bit
16 bit
[waste 16 bits to align the next item]
32 bit

whereas if it is re-arranged into

64 bit
64 bit
32 bit
32 bit
16 bit
[waste 16 or 48 bits here to get the entire object to be a multiple of 32 or 64 bits so that it can be used in arrays]

you can save some space on the padding. Mostly an issue for things that you have A LOT of, or if the members are laid out pathologically bad. (i.e. char, pointer, char, pointer, char, pointer, ...)

seanmakesgames: Yep. I'm familiar.
Luckily, EvalParams are not created in any significant manner. (1 per script run)
Unluckily for you, you're just going have to deal with knowing that it's staying there in the code. ;)

BrianHawley:
Probably best to add .DS_Store and .vscode to the .gitignore list, rather than add them here.

seanmakesgames:
One of my players is working on a new PR with updated v8 apis. Closing this one.

PR #64, opened 2017-08-22
Use the default upsteram implementation of an ArrayBufferAllocator

ignisf:
A change in upstream V8 was causing loading mini_racer to fail with a missing
symbol error. Switching to the default allocator provided by V8 works around
this issue and also simplifies the code base.

Fixes: https://github.com/cowboyd/libv8/issues/245

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -11,20 +11,6 @@
 
 using namespace v8;
 
-class ArrayBufferAllocator : public v8::ArrayBuffer::Allocator {
- public:
-  virtual void* Allocate(size_t length) {
-    void* data = AllocateUninitialized(length);
-    return data == NULL ? data : memset(data, 0, length);
-  }
-  virtual void* AllocateUninitialized(size_t length) {
-    return malloc(length);
-  }
-  virtual void Free(void* data, size_t) {
-     free(data);
-  }
-};
-
 typedef struct {
     const char* data;
     int raw_size;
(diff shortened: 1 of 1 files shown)


ignisf:
For reference, this is it in 6.1:

```c++
class ArrayBufferAllocator : public v8::ArrayBuffer::Allocator {
 public:
  virtual void* Allocate(size_t length) {
    void* data = AllocateUninitialized(length);
    return data == NULL ? data : memset(data, 0, length);
  }
  virtual void* AllocateUninitialized(size_t length) { return malloc(length); }
  virtual void Free(void* data, size_t) { free(data); }

  virtual void* Reserve(size_t length) {
    return base::VirtualMemory::ReserveRegion(length);
  }

  virtual void Free(void* data, size_t length,
                    v8::ArrayBuffer::Allocator::AllocationMode mode) {
    switch (mode) {
      case v8::ArrayBuffer::Allocator::AllocationMode::kNormal: {
        return Free(data, length);
      }
      case v8::ArrayBuffer::Allocator::AllocationMode::kReservation: {
        base::VirtualMemory::ReleaseRegion(data, length);
        return;
      }
    }
  }

  virtual void SetProtection(
      void* data, size_t length,
      v8::ArrayBuffer::Allocator::Protection protection) {
    switch (protection) {
      case v8::ArrayBuffer::Allocator::Protection::kNoAccess: {
        base::VirtualMemory::UncommitRegion(data, length);
        return;
      }
      case v8::ArrayBuffer::Allocator::Protection::kReadWrite: {
        const bool is_executable = false;
        base::VirtualMemory::CommitRegion(data, length, is_executable);
        return;
      }
    }
  }
};
```

SamSaffron:
I really like this, going to push out a new version shortly!

PR #68, opened 2017-08-24
fix minitest deprecation notices

seanmakesgames:
DEPRECATED: Use assert_nil if expecting nil from /Users/user/mini_racer/test/mini_racer_test.rb:32. This will fail in Minitest 6.


SamSaffron:
seems super fine to me!

PR #70, opened 2017-09-05
Add gvl reacquisition crash test case and resolution

### LINKED ISSUE #69: GVL lock issue with recursive attached proc calls
`rb_thread_call_with_gvl: called by a thread which has GVL`

Here's a test case:
```
  def test_attached_recursion
    context = MiniRacer::Context.new(timeout: 20)
    context.attach("a", proc{|a| a})
    context.attach("b", proc{|a| a})

    context.eval('const obj = {get r(){ b() }}; a(obj);')
  end
```

Ideally this would be some sort of exception instead of a c-level crash. What do you think @SamSaffron?

n00bish:
Closes #69

When running attached code, gvl is acquired if it is not already obtained, but the isolate is not informed of this state change.  I changed the isolate GetData/SetData to use an enum for clarity, and wrapped the gvl-acquiring callback in what should be the proper state manipulation logic to prevent the crash as described in the issue linked above.  Test case has been added and is passing with the code change.

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -56,6 +56,13 @@ typedef struct {
     long max_memory;
 } EvalParams;
 
+enum IsolateFlags {
+    IN_GVL,
+    DO_TERMINATE,
+    MEM_SOFTLIMIT_VALUE,
+    MEM_SOFTLIMIT_REACHED,
+};
+
 static VALUE rb_eScriptTerminatedError;
 static VALUE rb_eV8OutOfMemoryError;
 static VALUE rb_eParseError;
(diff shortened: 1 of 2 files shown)


seanmakesgames:
@SamSaffron bonus stuff in this PR: Had n00bish add enums for the numeric isolate data to make it more readable. Let us know if the code style is not as expected.

n00bish:
@SamSaffron quick reminder this is pending - should be a simple change, if you can spare a few minutes.  Any idea when you'd be able to take a look?

SamSaffron:
Sorry about this, I have been out on holidays for 2 weeks and this week I am at the team Discourse meetup. 

I really like this PR lets merge it in!

PR #79, opened 2018-03-13
Add type checking to user callable APIs

aviat:
This makes the API just a bit more user friendly, preventing segfaults when some functions are called with incorrect argument types.


SamSaffron:
awesome, totally support this!

PR #80, opened 2018-03-14
Add a call helper to call JS functions in a more performance way

aviat:
# Goal

Allow to call JS functions directly.

# Performance

This improve performance of calling functions directly, rather than passing them through `eval`:
```
Warming up --------------------------------------
                eval     9.062k i/100ms
                call    12.708k i/100ms
Calculating -------------------------------------
                eval     99.386k (± 7.2%) i/s -    498.410k in   5.042353s
                call    138.059k (± 4.8%) i/s -    698.940k in   5.074396s
```

# Limitations

No support for:

 - memory limitation
 - complex error processing


SamSaffron:
Nice, this looks really good to me!  I am merging it in! 

I want to do a release but would really love https://github.com/discourse/mini_racer/issues/75 fixed first, any chance you can have a look?

aviat:
Hey @SamSaffron not at the moment I'm afraid :(
Our next steps would probably [target this](https://github.com/discourse/mini_racer/blob/47329ae9c9a1dc99ab38fe2d701576a2864400d7/ext/mini_racer_extension/mini_racer_extension.cc#L326) which has a major performance impact when `eval` or `call` return objects.

SamSaffron:
sure, anything to make mini racer faster is totally welcome!

PR #87, opened 2018-05-09
Fix premature destruction of isolates

cataphract:
Fixes https://github.com/discourse/mini_racer/issues/75

I tried to maintain the optimization that caused isolates that are used by only one context to be also disposed of once the context is disposed of (by calling `dispose()`).

Unfortunately, as soon as a ruby `MiniRacer::Isolate` object is created, we need to hold an extra reference to a `v8::Isolate`. Because ruby has no reference counting, the `MiniRacer::Isolate` will only be reaped once GC runs.

To work around this problem, I avoid creating a  `MiniRacer::Isolate` as much as possible. Such an object is only created if `MiniRacer::Isolate.isolate` happens to be called. This also required taking the isolate mutex (the ruby one, which is used to avoid deadlocks between the GVL and `v8::Locker`) outside of `MiniRacer::Isolate` (otherwise the `Context` methods calling `isolate.with_lock` would bring the `Isolate` object into existence).

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -7,6 +7,7 @@
 #include <pthread.h>
 #include <unistd.h>
 #include <mutex>
+#include <atomic>
 #include <math.h>
 
 using namespace v8;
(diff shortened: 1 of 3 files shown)


aviat:
Thanks a lot @cataphract! @SamSaffron, @cataphract is a close colleague. Let us know if you need anymore information here!

SamSaffron:
This is great, thank you so much, will do a release with this as soon as @ignisf  is done upgrading libv8

PR #88, opened 2018-05-15
Improvements on returning values

cataphract:
* Do not create new contexts in `convert_v8_to_ruby` (this function is currently very inefficient because it even creates new contexts on recursive calls).
* Use the same logic in `eval` and `call` for transforming the result. This will give preference to converting the result to json in most situations, rather than using `convert_v8_to_ruby`.


SamSaffron:
I love this! thanks!

PR #94, opened 2018-07-31
patch up nil crash with symbols

seanmakesgames:
See test case. This is a quick fix to a segfault. Comments / feedback welcome. :)

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -320,6 +320,11 @@ static VALUE convert_v8_to_ruby(Isolate* isolate, Handle<Value> &value) {
         return rb_time_new(secs, nanos);
     }
 
+    if (value->IsSymbol())
+    {
+        return Qnil;
+    }
+
     if (value->IsObject()) {
 	VALUE rb_hash = rb_hash_new();
 	TryCatch trycatch(isolate);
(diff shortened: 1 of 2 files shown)


ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -320,6 +320,11 @@ static VALUE convert_v8_to_ruby(Isolate* isolate, Handle<Value> &value) {
         return rb_time_new(secs, nanos);
     }
 
+    if (value->IsSymbol())
+    {
+        return Qnil;
```
SamSaffron: shouldn't we translate this to a ruby symbol? Symbol("foo") => :foo ?

ignisf: This is what I was asking myself, too.

seanmakesgames: We definitely could. I didn't have time when I put this together to add that feature. (and I tried a bit with constructing / flattening the symbol to a string, and that didn't work out)

Happy to take any implementations for this from you. I could also add a thing to the readme 'js symbols are not supported at this time and return nil across procs'

Let me know what you think.

SamSaffron: I would very much prefer to fix this than document it as "it was too complicated for us to build :) " I can give it a shot over the next week or 2 unless someone beats me to it.

SamSaffron:
I just fixed this :confetti_ball:  doing a new release shortly

PR #97, opened 2018-08-13
Isolation/Vendoring

cataphract:


ignisf:
Hello, @cataphract,

Are you sure you opened this PR for the correct repository? This seems like it's better suited for a private fork of Sqreen? @aviat?

cataphract:
Sorry. I obviously didn't. In fact, the point of this is essentially to allow coexistence with projects that already use mini_racer and therubyracer, while ensuring we use a specific version of mini_racer.

PR #100, opened 2018-08-22
Add ability to dump snapshot to string

ianks:
In the future, I would like to be able to dump snapshot to a file and then load the binary snapshot file in a different process.  In order to do that, first we need to be able to dump the snapshot so it can be saved somewhere. `Snapshot#dump` is the first step in making that happen.  `Snapshot#dump` returns an ASCII-8BIT string which can then be saved off.

My next step after this is merged will be to work on loading this snapshot binary. Let me know if you see any issues, still learning the Ruby C api :)

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -562,6 +562,13 @@ static VALUE rb_snapshot_load(VALUE self, VALUE str) {
     return Qnil;
 }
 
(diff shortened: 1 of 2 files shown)


ianks:
I dont think the CI error was caused by my changes, but can someone double check??

SamSaffron:
yeah failure does not look related. there is an internal issue that needs debugging. 

Merging this in, seems fine to me.

PR #101, opened 2018-08-23
Add ability to load snapshots from a blob

ianks:
This is a follow up for https://github.com/discourse/mini_racer/pull/100

This PR adds the ability to load a snapshot from a binary. I have tested this a decent bit, and have not found and memory leaks, but it would be nice to get another set of eyes on it :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,21 @@ context.eval('counter')
 
 ```
 
+#### Dumping snapshots
+
+It is also possible to dump a `MiniRacer::Snapshot` and reuse is across isolated processes. This makes it so you only have to pay for the performance of creating a snapshot once. For example, you could build a snapshot of your JavaScript code in CI, then use it on your production servers.
+
+```ruby
+snapshot = MiniRacer::Snapshot.new('function renderToString() { return "<span>Hey!</span>"; }')
+File.write('my_snapshot.bin', snapshot.dump) # could also be a Redis cache :)
+
+restored_snapshot = MiniRacer::Snapshot.from_blob(File.read('my_snapshot.bin'))
+context = MiniRacer::Context.new(snapshot: snapshot)
+
+context.eval('renderToString()')
+# => "<span>Hey!</span>"
+```
+
 ### Shared isolates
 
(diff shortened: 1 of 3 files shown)


ignisf:
(it also fails to build on macos)

SamSaffron:
validating a snapshot sounds like a very hairy problem, have you tried discussing this on the v8 mailing list? Can you post something and link it here?

SamSaffron:
@ianks any news here?

ignisf:
is it necessary to work with the raw snapshot? This task could be simplified by wrapping the snapshot in a container with some metadata and ser/deserializing the whole thing

ignisf:
This [stack overflow answer](https://stackoverflow.com/a/46774933) makes me think that this may be an OK approach:

> (V8 developer here.) No, the format of V8's snapshot files is an internal implementation detail that is neither documented nor assumed to be stable across versions (on the contrary; V8 assumes that any snapshot that wasn't created by the exact same version is incompatible). There is no supported way to load a snapshot file into a disassembler; the way to inspect a snapshot's contents is to debug its creation and/or deserialization.

ianks:
@ignisf I was hoping to see some movement on the chromium side, but they have not created public api for checking the version. i was somewhat hesitant to create our own snapshot blob versioning as it’s incompatible with other libv8, and probably less robust than v8s, but it may be the best initial approach. let’s fogure out a serialization format for this, maybe just a 128 byte header with the version of v8?

nightpool:
Now that we're using the Node fork, does it have utilities for snapshot versioning we can use? Or maybe we could just use `LibV8::Node::VERSION`? Also, would it be possible to merge this PR anyway without worrying about versioning, and then add versioning as a "nice to have" feature later?

SamSaffron:
I like the idea of adding a prefix to the file with the version being LibV8::Node::VERSION ... then we can validate against it and only ever support 100% parity in versions.

ianks:
Hey sorry, I'm no longer using mini_racer at my new gig so don't have time to finish this up

PR #105, opened 2018-09-25
Remove conditional logic and use just libc++ on darwin irregardless of version

ignisf:
This fixes https://github.com/discourse/mini_racer/issues/104
Also fixes https://github.com/discourse/mini_racer/issues/107

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+- Unreleased
+
+  - Drop macOS feature detection and always link against libc++ macOS
+
 - 26-09-2018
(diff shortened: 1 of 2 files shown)


SamSaffron:
I think if we detect Darwin we should run this

https://unix.stackexchange.com/a/712/207

Pretty cheap and will give version for reals

For bonus points we can write a warning to stderr and complain that ruby is
not telling the truth anymore

SamSaffron:
OK just pushed a fix for this.

PR #106, opened 2018-09-26
Merge up 26 sep 2018

cataphract:


cataphract:
Sorry, wrong target again :)  I'll change the defaults so this don't happen again.

PR #116, opened 2018-11-19
Remove uses of deprecated non-isolate version of String::Utf8Value

ignisf:


SamSaffron:
I see, seems good to me! thanks!

PR #117, opened 2018-11-19
Remove use of non-maybe version of Message functions

ignisf:
I think we need to check if these methods return Nothing. @SamSaffron do you have any preference what the printed output should be in case the column or the line are undefined?


SamSaffron:
oh no, I made this conflict, did not merge in correct order :(

ignisf:
Take a look at the second commit, I think it makes more sense like this.

SamSaffron:
I am so on the fence here, feel free to pick whichever you prefer and commit. Kind of leaning towards doing the tochecked thing

ignisf:
I am a bit concerned that it would introduce segfaults during  printing an exception... Which is unpleasant

SamSaffron:
I am merging this in, we need something here and this is holding up 7.3 which is urgent.

PR #118, opened 2018-11-19
Remove use of non-maybe version of TryCatch::StackTrace

ignisf:


SamSaffron:
looks good to me ... technically we need to check if we have a string I think if we do a to local checked, no?

ignisf:
I think we both check if the StackTrace is empty and then do a ToLocalChecked before getting the string so we should not have a crash.

PR #124, opened 2018-12-21
Fix memory leak in gc_callback

messense:


SamSaffron:
ouch ... this looks right to me, thanks!

PR #125, opened 2019-01-04
Lock libv8 to < 7.0

ianks:
Currently, there seem to be compilation issues when attempting to compile with
lib8 v7. This patch ensures we only attempt to build the gem against a
compatible v6 version of libv8.


SamSaffron:
@ignisf  do we have a topic to track issues with libv8 7 ? are there any know incompatibilities ?

ignisf:
There's an issue (https://github.com/cowboyd/libv8/issues/274), however I've not released a stable version of libv8 7

PR #127, opened 2019-01-26
Unify indentation style

Taiki-San:
Currently, the most common indentation style of the C++ file is 4 spaces per level.
However, tabs are sometimes used assuming an 8 spaces width (where a single tab is used in place of two space-based indentation levels).
Although there is no difference on GitHub (where tabs are 8-wide), it results in an extremely confusing appearance when the tab width is configured differently. Specifically, when tabs are configured at 4 spaces (which is a common default), and the additional indentation level simply disappear.

This PR replace any tabs in `mini_racer_extension.cc` by 8 spaces making the code always render as it is meant to.

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -71,7 +71,7 @@ public:
     }
 
     int refs() {
-	return refs_count;
+        return refs_count;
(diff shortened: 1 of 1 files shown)


SamSaffron:
OK ... I guess we got to clean this up... thank Taiki

PR #130, opened 2019-02-28
Add example of using source-map-support

ianks:
This PR adds an example for setting up robust source map support for webpack. Happy to make any changes to make this more useful :)


SamSaffron:
Cool, some minor changes needed:

1. Can you link to the example from the main readme?
2. Can you ensure the gemspec bypasses adding the examples directory

Anyone here feel like trying this out?

ianks:
@SamSaffron addressed your comments :smile:

SamSaffron:
Nice, thanks @ianks !

PR #133, opened 2019-03-13
Fix install on ruby230 macos1014

lloeki:
We started including `ruby.h` between 0.2.4 sqreen1 and sqreen2, which in turns includes `missing.h` which on 2.3.0 has the following:

```
#ifndef HAVE_EXPLICIT_BZERO
RUBY_EXTERN void explicit_bzero(void *b, size_t len);
# ifdef HAVE_MEMSET_S
# include <string.h>
static inline void
explicit_bzero_by_memset_s(void *b, size_t len)
{
    memset_s(b, len, 0, len);
}
#   define explicit_bzero(b, len) explicit_bzero_by_memset_s(b, len)
# elif defined SecureZeroMemory
#   define explicit_bzero(b, len) SecureZeroMemory(b, len)
# endif
#endif
```

Somehow `memset_s` detection is incorrect. Forcing it with `# undef HAVE_MEMSET_S`.


lloeki:
Sorry for the noise. Incorrect repo target.

PR #139, opened 2019-04-24
Introduce a rake lint task that runs clang-tidy on the source

ignisf:


SamSaffron:
I am fine with this, we should probably clean stuff up

PR #141, opened 2019-05-17
Introduce a rake test:asan task for convenient running of Address Sanitiser

ignisf:
Sample output:

image


ignisf:
P.S. 
The detected error is in the last part of the conditional statement here:

https://github.com/discourse/mini_racer/blob/f407f68b534c16193222e9365841ac22bcc2ba4b/ext/mini_racer_extension/mini_racer_extension.cc#L1214-L1216

I am not too sure what's the idea here, so I chose not to change it.

SamSaffron:
I am more than happy to try this out!

PR #142, opened 2019-05-28
Compatibility changes for V8 7.4 support

ignisf:


ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -152,7 +152,7 @@ static VALUE rb_mJSON;
 static VALUE rb_cFailedV8Conversion;
 static VALUE rb_cDateTime = Qnil;
 
-static Platform* current_platform = NULL;
+static std::unique_ptr<Platform> current_platform = NULL;
```
ignisf: @SamSaffron, check this out. The new platform API mandates the use of an `sstd::unique_ptr`.

SamSaffron: oh interesting!

SamSaffron:
OK lets try this out! CI is freaking out about bundler :( will get it sorted

ignisf:
They should be releasing a new version of bundler that no longer has this issue.

Issue #160, opened 2020-04-08
The call method, through nogvl_context_call, does not seem to enforce the max_memory limit


masongup-mdsol:
I am still investigating what's happening here, but it appears that the `max_memory` limit is only enforced when running Javascript code through the `exec` method - see [the code setting the data point on the isolate and the `AddGCEpilogueCallback` here](https://github.com/rubyjs/mini_racer/blob/master/ext/mini_racer_extension/mini_racer_extension.cc#L352-L355), in the `nogvl_context_eval` method. No comparable code seems to be present in the `nogvl_context_call` method.

The code that I am currently working with loads a few hundred lines of JS via the `eval` method which defines various Javascript functions, and then calls the defined methods with Ruby objects using the `call` method. I was hoping to use the `max_memory` limit in this code to prevent excessive memory consumption, but it does not seem to be effective. I added a log line to the C callback function `gc_callback`, and the memory limits I set appear to get stomped by the time I run any of the defined functions - when it runs, it thinks that the softlimit is some random huge value.

At this stage, what I am asking is whether there is an intentional reason for this. If this is not an intended behavior, I am open to creating a test exercising this, and writing a fix.

SamSaffron:
I don't think this is intentional, would be more than happy to see a fix.

masongup-mdsol:
I've been looking into this some more. Based on what I've seen, it seems that something is overwriting the value for the `MEM_SOFTLIMIT_VALUE` in the isolate under some conditions. When I tried to write a Ruby test for it in mini_racer, it was overwritten to `0`, and would exit on the first GC run. When I was running the original code in the application I'm working, it would get overwritten to a very high value, and would never trigger. I see 2 possible ways to address this issue:

1. Figure out and fix whatever is overwriting the `MEM_SOFTLIMIT_VALUE`. I am not an expert in C++, and probably don't have bandwidth to figure that out.

2. Add a bit of code in `nogvl_context_call` to set the memory limit if present, just like is already present in `nogvl_context_eval`. I have implemented this locally and tested it, and it seems to work as expected, both in the test I wrote on mini_racer, and in our application. I'll open a PR with this change.

One other note - it doesn't seem to be possible to write a Ruby test exercising this issue. The memory limiter still runs with or without the fix, it just runs with an incorrect value for `MEM_SOFTLIMIT_VALUE`, and thus terminates at the wrong time.

masongup-mdsol:
[...]

PR #161, opened 2020-04-10
Set memory softlimit in nogvl_context_call also

masongup-mdsol:
This PR addresses issue  #160, where the memory limit was not being enforced when the `call` method is used to execute Javascript functions. It explicitly sets the `MEM_SOFTLIMIT_VALUE` and `MEM_SOFTLIMIT_REACHED` values each time `call` is used, as is already done in the `eval` code. I have observed the `MEM_SOFTLIMIT_VALUE` being overwritten with seemingly random data between Ruby calls, causing the memory limiter to trip either too soon or not at all.

I am not sure if it would be practical to write a Ruby test for this, as we would have to assert that the code terminates with a `V8OutOfMemoryError` at a specific amount of memory used, and not before or after.

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -127,6 +127,7 @@ typedef struct {
     Local<Function> fun;
     Local<Value> *argv;
     EvalResult result;
+    size_t max_memory;
 } FunctionCall;
 
 enum IsolateFlags {
@@ -1454,6 +1455,12 @@ nogvl_context_call(void *args) {
     // terminate ASAP
     isolate->SetData(DO_TERMINATE, (void*)false);
 
+    if (call->max_memory > 0) {
+        isolate->SetData(MEM_SOFTLIMIT_VALUE, &call->max_memory);
+        isolate->SetData(MEM_SOFTLIMIT_REACHED, (void*)false);
+        isolate->AddGCEpilogueCallback(gc_callback);
+    }
+
     Isolate::Scope isolate_scope(isolate);
     EscapableHandleScope handle_scope(isolate);
     TryCatch trycatch(isolate);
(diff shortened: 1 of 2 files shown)


SamSaffron:
Change looks safe to me... thanks! 

Yeah a test is hard here, but yeah maybe if we set the limit low enough and allocate a giant string we can trigger it reliably? 

Can you give it a shot, if too hard I can just merge. Key is test needs to be both fast and reliable.

masongup-mdsol:
I added a test that passes with my change on my system, and fails if I comment out line 1459.

The tricky part isn't that the memory limiter fails to trigger, but that the limit value gets overwritten. I had to add the `set_s` and `context.call('set_s', 1000)` code to get it to trigger. On my system, that overwrites the memory limit to zero, causing it to trigger immediately. I can catch that this happened by asserting that the context's `s` is sufficiently high, so the loop hasn't run a number of times that's consistent with actually hitting the memory limit.

I have also observed the memory limit getting overwritten to a massively high value in the codebase for our project that I originally found this issue in. I am not sure how to trigger that to happen reliably, but I was concerned that it might happen on some other system configuration. If that happens, then the test suite segfaults after that test takes up all of the available memory. To prevent this, I added a bit of Javascript code to exit cleanly if the value gets too high. The test catches this by asserting that the exception was raised, so it fails if this break causes it to exit cleanly.

I think these checks effectively assert that the memory limit is there, and at least roughly in the range that it was explicitly set to. I don't know how the overwrite is happening exactly, but I'm assuming that it's very unlikely that it would be overwritten to a value that would allow this test to pass.

SamSaffron:
Looks good to me!

masongup-mdsol:
Excellent, thanks @SamSaffron! Any idea when this can be released to Rubygems in an updated version?

SamSaffron:
No probs, released a new version just now.

PR #153, opened 2019-12-05
WIP Mechanical compatibility changes

ignisf:

PR #162, opened 2020-05-05
Fix a use-after-free and a shutdown crash

cataphract:
See the commit messages for more details.

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -156,6 +156,10 @@ static VALUE rb_cDateTime = Qnil;
 static std::unique_ptr<Platform> current_platform = NULL;
 static std::mutex platform_lock;
 
+static pthread_attr_t *thread_attr_p;
+static pthread_rwlock_t exit_lock = PTHREAD_RWLOCK_INITIALIZER;
+static bool ruby_exiting; // guarded by exit_lock
+
 static VALUE rb_platform_set_flag_as_str(VALUE _klass, VALUE flag_as_str) {
     bool platform_already_initialized = false;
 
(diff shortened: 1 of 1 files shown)


cataphract:
I forgot, this also fixes a leak; the threads created to destroy contexts were not detached, so they were expected to be joined.

ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -1235,22 +1253,17 @@ static void free_context(ContextInfo* context_info) {
...
     }
 
-    if (context_info->context && isolate_info && isolate_info->isolate) {
-        context_info->context = NULL;
-    }
-
-    if (isolate_info) {
-        context_info->isolate_info = NULL;
-    }
+    context_info->context = NULL;
```
SamSaffron, line 1265: just confirming we don't need the guards anymore?

SamSaffron:
@cataphract  how would you feel about working on getting mini racer working with latest libv8 

We are struggling here: https://github.com/rubyjs/mini_racer/pull/153#issuecomment-623561107

ext/mini_racer_extension/mini_racer_extension.cc
cataphract: Yes, the problem is that calling `dispose`  on the context triggers the `IsolateInfo` to be deleted (through `free_context_raw`), but afterwards we attempt to access `context_info->isolate_info->isolate`.

SamSaffron:
cool, lets try this out. Thanks!

SamSaffron:
One big issue with this implementation is that it is fork unsafe. I can reproduce hangs on local with forked processes. 

This is particularity a big issue with my experimentation around single threaded platform: 

https://github.com/rubyjs/mini_racer/tree/single_threaded_platform

we are going to need to figure out a way of making this fork safe.

cataphract:
@SamSaffron can you provide the code to reproduce it? Is it the problem that the mutex is in a bad state after the fork?

SamSaffron:
Sure @cataphract give a shot in https://github.com/rubyjs/mini_racer/tree/single_threaded_platform

specifically test/test_forking.rb hangs for me. 

The unfortunate thing here though is that you are going to have to build a custom v8 for this to hang on exit https://github.com/rubyjs/mini_racer/issues/175#issuecomment-705898043

cataphract:
@SamSaffron ok, I'll look at it this week

SamSaffron:
thanks @cataphract !

cataphract:
@SamSaffron I've added a PR: https://github.com/rubyjs/mini_racer/pull/180

PR #163, opened 2020-05-09
Use (unreleased) libv8 8.4.307.0

cataphract:
There are some new warnings:

```
../../../../ext/mini_racer_extension/mini_racer_extension.cc:1163:58: warning: ignoring return value of ‘v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)’, declared with attribute warn_unused_result [-Wunused-result]
                                         .ToLocalChecked());
```

These are due to ignoring the return value of `->Set()`. I'm not sure if ignoring this return value (with a pending exception) is an acceptable state of affairs, so I let them be.

I've only tested libv8 and mini_racer on linux amd64

### DIFF
--- a/ext/mini_racer_extension/extconf.rb
+++ b/ext/mini_racer_extension/extconf.rb
@@ -11,6 +11,7 @@ $CPPFLAGS += " -rdynamic" unless $CPPFLAGS.split.include? "-rdynamic"
 $CPPFLAGS += " -fPIC" unless $CPPFLAGS.split.include? "-rdynamic" or IS_DARWIN
 $CPPFLAGS += " -std=c++0x"
 $CPPFLAGS += " -fpermissive"
+$CPPFLAGS += " -DV8_COMPRESS_POINTERS"
 
 $CPPFLAGS += " -Wno-reserved-user-defined-literal" if IS_DARWIN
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Amazing work @cataphract  

@ignisf  any chance we can release a new libv8? We can not merge this in without it being out there

nightpool:
@cataphract Not sure if you saw @bnoordhuis' comments on the other PR about the warmup issue:

> If you have an existing snapshot when creating the new one, pass it as the third argument to SnapshotCreator, that should ensure they use the same read-only heap.
>
> If you don't pass it, it's going to use the default snapshot (if you use V8::SetSnapshotBlob() - but it looks like you don't) or no snapshot.
>
> Compiling without RO heap is fine too. It's an optimization (less work for the GC) but not a critical one.

cataphract:
> @cataphract Not sure if you saw @bnoordhuis' comments on the other PR about the warmup issue:
> 
> > If you have an existing snapshot when creating the new one, pass it as the third argument to SnapshotCreator, that should ensure they use the same read-only heap.
> > If you don't pass it, it's going to use the default snapshot (if you use V8::SetSnapshotBlob() - but it looks like you don't) or no snapshot.
> > Compiling without RO heap is fine too. It's an optimization (less work for the GC) but not a critical one.

Yeah but v8 defaults to having the RO heap disabled, at least on Linux, so I let it be.

SamSaffron:
@cataphract I am fine with that, I guess the blocker here is getting a new libv8 released

SamSaffron:
@cataphract thanks heaps for your work here, we have a new release now with 8.4!!!

PR #166, opened 2020-07-15
Upgrade to libv8 8.4

nightpool:
Based on #163 with a slight tweak to the gemspec and tests. Requires https://github.com/rubyjs/libv8/pull/301/ to be merged.

Future versions of v8 (8.4 stable, 8.5 beta, and 8.6 development) are blocked on https://bugs.chromium.org/p/v8/issues/detail?id=10041. Right now you can test locally by installing https://github.com/nightpool/libv8/packages/314620, but be aware that I've only pushed a version for `x86_64-linux`, since that's what I was testing on (specifically, i tested on the `ruby:2.5.8-stretch` docker container).

`rake test` passes locally, I haven't tried any of the other tests yet.

closes #153, #163

### DIFF
--- a/ext/mini_racer_extension/extconf.rb
+++ b/ext/mini_racer_extension/extconf.rb
@@ -11,6 +11,7 @@ $CPPFLAGS += " -rdynamic" unless $CPPFLAGS.split.include? "-rdynamic"
 $CPPFLAGS += " -fPIC" unless $CPPFLAGS.split.include? "-rdynamic" or IS_DARWIN
 $CPPFLAGS += " -std=c++0x"
 $CPPFLAGS += " -fpermissive"
+$CPPFLAGS += " -DV8_COMPRESS_POINTERS"
 
 $CPPFLAGS += " -Wno-reserved-user-defined-literal" if IS_DARWIN
 
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -25,6 +25,7 @@ public:
     ArrayBuffer::Allocator* allocator;
     StartupData* startup_data;
     bool interrupted;
+    bool added_gc_cb;
     pid_t pid;
     VALUE mutex;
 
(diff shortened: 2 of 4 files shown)


SamSaffron:
This is fantastic @nightpool  ... @ignisf  can we merge this?

nightpool:
Unfortunately i'm still having problems with my React SSR related to locale data. I'm working on a further patch for libv8, but my current working theory is that libv8 isn't packaging locale data at all. Adding it statically to the v8 `so` looks pretty easy, but maybe it should do so as a separate gem or an optional configuration?

SamSaffron:
I would say it depends on how much extra size would be added by adding locale data?

nightpool:
Okay, it looks like the locale data isn't _too_ much bigger—my libv8 .gem went from 11.9 MB to 16.2 MB

SamSaffron:
Will leave this for @ignisf to decide but my gut feel is that it is OK. 

Is there are way to package it so it is an additional dependency eg: `libv8-locale` ?

nightpool:
I haven't looked into it yet. I'm sure it's possible, not sure what the additional complexity would be or how long it would take to get working.

nightpool:
I'm going to push two more PRs today to trigger travis tests with my pre-release libv8 packages, just so we can make sure tests are working on the full suite of platforms as well as locally for me.

SamSaffron:
@nightpool shall I merge this now? is is backwards compat?

nightpool:
@SamSaffron I haven't investigated the c++ changes in detail, because I was mostly building on the work @cataphract and @ignisf did, but my guess is probably not. (especially since I bumped the gemfile to require libv8 8.4, which hasn't been released yet)

SamSaffron:
K ... going to hold off on merging this till we get libv8 8.4 published.

nightpool:
@SamSaffron @cowboyd @ignisf built libv8 gems are ready at https://github.com/rubyjs/libv8/releases/tag/v8.4.255.0, I've tested them locally and we're just waiting on getting them pushed to rubygems before we can merge this!

SamSaffron:
Amazing work @nightpool I just pushed a release to rubygems! merging this in!

nightpool:
yes!!!!!!!! so great to get this merged.

PR #167, opened 2020-07-16
DEV: EOL Ruby 2.4 is no longer officially supported

runlevel5:


SamSaffron:
Sure this is fine with me!

PR #177, opened 2020-10-02
[spike] Implement setTimeout

ianks:
This PR is a spike at adding an optional `setTimeout` implementation. The current code is very raw, but wanted to post is so I could get some early feedback. Thoughts?

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -188,6 +188,14 @@ static VALUE rb_platform_set_flag_as_str(VALUE _klass, VALUE flag_as_str) {
     return Qnil;
 }
 
+typedef struct {
+    Persistent<Function, CopyablePersistentTraits<Function>>* v8_func;
+    // Persistent<Function>* v8_func;
+    Persistent<Context>* context;
+    Isolate* isolate;
+} JavascriptFunctionInfo;
+
+
 static void init_v8() {
     // no need to wait for the lock if already initialized
     if (current_platform != NULL) return;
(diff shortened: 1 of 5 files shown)


SamSaffron:
We have to be super careful around lifetimes here between the 2 vMs we don't want to segfault cause a context with timeouts got disposed. 

I can see people wanting both this and setInterval though I wonder if this scheduling should happen directly in C to avoid depending on Ruby here 

Do you have a particular use case here you are blocked on. 

Maybe a good opening move here is simply offering some level of interoperability in general? 

```
fn = ctx.eval("function(x) {return 40 + x;}")
ctx.call(fn, 2)
```

Ideally the amount of info we leak between VMs though should be minimal. So stuff like this should not segfault or lead to any surprises. 

```
fn = ctx.eval("function(x) {return 40 + x;}")
ctx.dispose
p fn
ctx.call(fn, 2)
```

ianks:
> I can see people wanting both this and setInterval though I wonder if this scheduling should happen directly in C to avoid depending on Ruby here

I was wondering this as well. It [seems like there is support in libplatform](https://chromium.googlesource.com/v8/v8/+/master/src/libplatform/default-platform.cc#203) for an event loop that does this type of thing, but the documentation is scarce. Overall, if it is possible, I do think it is a better approach.

Here are some examples of how others have implemented timers:
- https://github.com/NativeScript/NativeScript/blob/master/packages/core/timer/index.ios.ts#L72
- https://github.com/NativeScript/NativeScript/blob/master/packages/core/timer/index.android.ts#L18
- https://github.com/chromium/chromium/blob/888beef69f57f6c8810d22af036e961208cda05f/third_party/blink/renderer/core/frame/window_or_worker_global_scope.cc#L132

If anyone stumble on any information about implementing this in an embedded context, please post it here! The documentation on it is hard to come by.

> Do you have a particular use case here you are blocked on.

We are SSR React in our codebase with GraphQL. Having the ability to make those queries async would speed things up a bunch.

> Maybe a good opening move here is simply offering some level of interoperability in general?
> 
> ```
> fn = ctx.eval("function(x) {return 40 + x;}")
> puts ctx.call(fn, 2)
> => 42
> ```

👍

ianks:
this task is beyond my current abilities for v8 and c++, so i won’t be able to implement. if anyone wants to take this on, have at it!!

SamSaffron:
@ianks I know it has been quite a few years, but our new design handles all of this now, we should be good.

Issue #175, opened 2020-09-13
MiniRacer hangs after fork


SamSaffron:
@cataphract  / @nightpool previous versions of mini_racer were fork friendly ... but now sadly it looks like we totally stall after forks due to a GC thread.

Repro is: 

```
  def test_forking
    trigger_gc = lambda do
      ctx = MiniRacer::Context.new
      ctx.eval("var a = #{('x' * 100000).inspect}")
      ctx.eval("a = undefined")
      ctx.isolate.low_memory_notification
    end
    trigger_gc.call
    Process.wait fork { trigger_gc.call }
  end
```

Any ideas how we can sort this out?

SamSaffron:
Posted on v8 users here: 

https://groups.google.com/g/v8-users/c/sosbtZ49m3U

SamSaffron:
Backtrace:

```
(gdb) bt
#0  0x00007f18d4fb202d in pthread_cond_signal@@GLIBC_2.3.2 () from /usr/lib/libpthread.so.0
#1  0x00007f18d084a5fc in v8::platform::DelayedTaskQueue::Append(std::unique_ptr<v8::Task, std::default_delete<v8::Task> >) ()
   from /home/sam/Source/mini_racer/lib/mini_racer_extension.so
#2  0x00007f18d0849db0 in v8::platform::DefaultWorkerThreadsTaskRunner::PostTask(std::unique_ptr<v8::Task, std::default_delete<v8::Task> >) () from /home/sam/Source/mini_racer/lib/mini_racer_extension.so
#3  0x00007f18d0848f51 in v8::platform::DefaultPlatform::CallOnWorkerThread(std::unique_ptr<v8::Task, std::default_delete<v8::Task> >) () from /home/sam/Source/mini_racer/lib/mini_racer_extension.so
#4  0x00007f18d1014653 in v8::internal::ConcurrentMarking::ScheduleTasks() ()
   from /home/sam/Source/mini_racer/lib/mini_racer_extension.so
#5  0x00007f18d09732ab in v8::internal::MarkCompactCollector::MarkLiveObjects() ()
   from /home/sam/Source/mini_racer/lib/mini_racer_extension.so
#6  0x00007f18d0972bb1 in v8::internal::MarkCompactCollector::CollectGarbage() ()
   from /home/sam/Source/mini_racer/lib/mini_racer_extension.so
[...]

PR #179, opened 2020-10-05
Isolate v8 symbols

lloeki:
Hello!

We identified an issue happening when two extensions are built with different V8 versions, then loaded at the same time.

In an ideal world, all extensions would be built with `-fvisibility=hidden`, only exposing relevant symbols and avoiding clases, unfortunately that is not the case.

Indeed the libv8 gem does not build its static library with `-fvisibility=hidden`. The result is that when this static library is linked into a Ruby extension (IOW a dynamic library), V8 symbols are global, irrespective of passing `-fvisibility=hidden` when building the extension.

Additionally, Ruby doesn't really make an effort to load its extensions in isolated ways, resulting in the default of globally available symbols.

As a result, the second extension being loaded will resolve some of its V8 symbols to the first ones that were loaded.

To combat this, within Sqreen's fork of MiniRacer, we created a small wrapper that takes in charge loading the extension more privately by:

- enforcing all symbols to be local (via `RTLD_LOCAL`), preventing other libs to pick its symbols up
- attempting to enforce binding to local symbols first (via `RTLD_DEEPBIND`), preventing this lib to pick others symbols

Unfortunately, the latter is not always available, resulting in MiniRacer's V8 symbols being picked up by other libraries, even when they use `RTLD_LOCAL` (which only controls things one way). In that case we are left with no option.

This PR attempts to make MiniRacer a good citizen in two ways:

- first commit: compile with `-fvisibility=hidden` to hide any irrelevant `mini_racer_extension` symbols, and be ready if/when `libv8_monolith.a` is compiled with it too.
- second commit: hide V8 symbols from `libv8_monolith.a` that wasn't built with `-fvisibility=hidden`.

Considered alternatives
- linker version scripts to hide symbols, but it's only supported by GCC's `ld`
- `-exported_symbols_list` and `-unexported_symbols_list`, but it needs exhaustive generation of symbols to hide and show. Possible, but calling into `nm libv8_monolith.a` and parsing the output to generate the list could be unreliable.
- leveraging gem dependencies to ensure the same `libv8` version *and* build: unfortunately this does not mean a given lib wasn't built with a different v8, unless all gems pin their `libv8` versions *exactly* in their `gemspec` (this particular one stumped us for a long time)

Excerpt of `mini_racer_extension.bundle` symbol table (`T` means global, `t` means local`):

```
$ nm -C lib/mini_racer_extension.bundle | grep ' T' | head
0000000000003ce0 T _Init_mini_racer_extension
00000000001544b0 T _v8_internal_Get_Object(void*)
0000000000154870 T _v8_internal_Node_Print(void*)
0000000000154550 T _v8_internal_Print_Code(void*)
00000000001544f0 T _v8_internal_Print_Object(void*)
00000000001547c0 T _v8_internal_Print_StackTrace()
0000000000154800 T _v8_internal_Print_TransitionTree(void*)
0000000000154730 T _v8_internal_Print_LayoutDescriptor(void*)
000000000000f260 T V8_Fatal(char const*, ...)
000000000000f3b0 T V8_Dcheck(char const*, int, char const*)
```

Excerpt of a stack trace where code jumps from one lib to the other *within libv8*:

```
-- C level backtrace information -------------------------------------------
/Users/lloeki/.rubies/ruby-2.7.1/bin/ruby(rb_vm_bugreport+0x96) [0x106ebf616]
/Users/lloeki/.rubies/ruby-2.7.1/bin/ruby(rb_bug_for_fatal_signal+0x1d0) [0x106cf9270]
/Users/lloeki/.rubies/ruby-2.7.1/bin/ruby(sigsegv+0x5b) [0x106e22c4b]
/usr/lib/system/libsystem_platform.dylib(_sigtramp+0x1d) [0x7fff7248c5fd]
/Users/lloeki/.gem/ruby/2.7.1/gems/mini_racer-0.3.1/lib/mini_racer_extension.bundle(_ZN2v88internal12_GLOBAL__N_123GetPageTableInitializerEv+0x25) [0x10bade405]
/Users/lloeki/.gem/ruby/2.7.1/gems/mini_racer-0.3.1/lib/mini_racer_extension.bundle(0x10badea39) [0x10badea39]
/Users/lloeki/.gem/ruby/2.7.1/gems/mini_racer-0.3.1/lib/mini_racer_extension.bundle(0x10b77b3b6) [0x10b77b3b6]
/Users/lloeki/.gem/ruby/2.7.1/gems/mini_racer-0.3.1/lib/mini_racer_extension.bundle(0x10b77bc5e) [0x10b77bc5e]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(_ZN2v88internal9SemiSpace6CommitEv+0x7a) [0x10d81df5a]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(0x10d81de76) [0x10d81de76]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(0x10d7a35ef) [0x10d7a35ef]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(0x10d74cf00) [0x10d74cf00]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(0x10d74c9fd) [0x10d74c9fd]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(0x10d5e78fb) [0x10d5e78fb]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(0x10d5e79af) [0x10d5e79af]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(_ZN11IsolateInfo4initEP12SnapshotInfo+0xc2) [0x10d3ba012]
/Users/lloeki/Workspace/github.com/sqreen/mini_racer/lib/sq_mini_racer_extension.bundle(_ZL22rb_context_init_unsafemmm+0x126) [0x10d3bc016]
/Users/lloeki/.rubies/ruby-2.7.1/bin/ruby(vm_call_cfunc+0x16c) [0x106eb0cdc]
/Users/lloeki/.rubies/ruby-2.7.1/bin/ruby(vm_exec_core+0x38b0) [0x106e97050]
```

### DIFF
--- a/ext/mini_racer_loader/mini_racer_loader.c
+++ b/ext/mini_racer_loader/mini_racer_loader.c
@@ -1,0 +1,123 @@
+#include <ruby.h>
+#include <dlfcn.h>
+#include <string.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+// Load a Ruby extension like Ruby does, only with flags that:
+// a) hide symbols from other extensions (RTLD_LOCAL)
+// b) bind symbols tightly (RTLD_DEEPBIND, when available)
+
+void Init_mini_racer_loader(void);
+
+static void *_dln_load(const char *file);
+
+static VALUE _load_shared_lib(VALUE self, volatile VALUE fname)
+{
+    (void) self;
+
+    // check that path is not tainted
+    SafeStringValue(fname);
+
+    FilePathValue(fname);
+    VALUE path = rb_str_encode_ospath(fname);
+
+    char *loc = StringValueCStr(path);
+    void *handle = _dln_load(loc);
+
+    return handle ? Qtrue : Qfalse;
+}
+
+// adapted from Ruby's dln.c
+#define INIT_FUNC_PREFIX ((char[]) {'I', 'n', 'i', 't', '_'})
+#define INIT_FUNCNAME(buf, file) do { \
+    const char *base = (file); \
+    const size_t flen = _init_funcname(&base); \
+    const size_t plen = sizeof(INIT_FUNC_PREFIX); \
+    char *const tmp = ALLOCA_N(char, plen + flen + 1); \
+    memcpy(tmp, INIT_FUNC_PREFIX, plen); \
+    memcpy(tmp+plen, base, flen); \
+    tmp[plen+flen] = '\0'; \
+    *(buf) = tmp; \
+} while(0)
+
+// adapted from Ruby's dln.c
+static size_t _init_funcname(const char **file)
+{
+    const char *p = *file,
+               *base,
+               *dot = NULL;
+
+    for (base = p; *p; p++) { /* Find position of last '/' */
+        if (*p == '.' && !dot) {
+            dot = p;
+        }
+        if (*p == '/') {
+            base = p + 1;
+            dot = NULL;
+        }
+    }
+    *file = base;
+    return (uintptr_t) ((dot ? dot : p) - base);
+}
+
+// adapted from Ruby's dln.c
+static void *_dln_load(const char *file)
+{
+    char *buf;
+    const char *error;
+#define DLN_ERROR() (error = dlerror(), strcpy(ALLOCA_N(char, strlen(error) + 1), error))
+
+    void *handle;
+    void (*init_fct)(void);
+
+    INIT_FUNCNAME(&buf, file);
+
+#ifndef RTLD_DEEPBIND
+# define RTLD_DEEPBIND 0
+#endif
+    /* Load file */
+    if ((handle = dlopen(file, RTLD_NOW|RTLD_LOCAL|RTLD_DEEPBIND)) == NULL) {
+        DLN_ERROR();
+        goto failed;
+    }
+#if defined(RUBY_EXPORT)
+    {
+        static const char incompatible[] = "incompatible library version";
+        void *ex = dlsym(handle, "ruby_xmalloc");
+        if (ex && ex != (void *) &ruby_xmalloc) {
+
+# if defined __APPLE__
+            /* dlclose() segfaults */
+            rb_fatal("%s - %s", incompatible, file);
+# else
+            dlclose(handle);
+            error = incompatible;
+            goto failed;
+#endif
+        }
+    }
+# endif
+
+    init_fct = (void (*)(void)) dlsym(handle, buf);
+    if (init_fct == NULL) {
+        error = DLN_ERROR();
+        dlclose(handle);
+        goto failed;
+    }
+
+    /* Call the init code */
+    (*init_fct)();
+
+    return handle;
+
+failed:
+    rb_raise(rb_eLoadError, "%s", error);
+}
+
+__attribute__((visibility("default"))) void Init_mini_racer_loader()
+{
+    VALUE mSqreen = rb_define_module("MiniRacer");
+    VALUE mPrvExtLoader = rb_define_module_under(mSqreen, "Loader");
+    rb_define_singleton_method(mPrvExtLoader, "load", _load_shared_lib, 1);
+}
--- a/ext/mini_racer_loader/mini_racer_loader.c
+++ b/ext/mini_racer_loader/mini_racer_loader.c
@@ -77,7 +77,7 @@ static void *_dln_load(const char *file)
 # define RTLD_DEEPBIND 0
 #endif
     /* Load file */
-    if ((handle = dlopen(file, RTLD_NOW|RTLD_LOCAL|RTLD_DEEPBIND)) == NULL) {
+    if ((handle = dlopen(file, RTLD_LAZY|RTLD_LOCAL|RTLD_DEEPBIND)) == NULL) {
         DLN_ERROR();
         goto failed;
     }
@@ -117,7 +117,7 @@ failed:
 
 __attribute__((visibility("default"))) void Init_mini_racer_loader()
 {
-    VALUE mSqreen = rb_define_module("MiniRacer");
-    VALUE mPrvExtLoader = rb_define_module_under(mSqreen, "Loader");
-    rb_define_singleton_method(mPrvExtLoader, "load", _load_shared_lib, 1);
+    VALUE mMiniRacer = rb_define_module("MiniRacer");
+    VALUE mLoader = rb_define_module_under(mMiniRacer, "Loader");
+    rb_define_singleton_method(mLoader, "load", _load_shared_lib, 1);
 }
--- a/lib/mini_racer.rb
+++ b/lib/mini_racer.rb
@@ -1,5 +1,12 @@
 require "mini_racer/version"
-require "mini_racer_extension"
+require "mini_racer_loader"
+
+ext_filename = "mini_racer_extension.#{RbConfig::CONFIG['DLEXT']}"
+ext_path = $LOAD_PATH.map { |p| (Pathname.new(p) + ext_filename) }.find { |p| p.file? }
+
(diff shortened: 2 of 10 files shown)


ext/mini_racer_loader/mini_racer_loader.c
```diff
@@ -0,0 +1,123 @@
...
+
+    return handle;
+
+failed:
+    rb_raise(rb_eLoadError, "%s", error);
+}
+
+__attribute__((visibility("default"))) void Init_mini_racer_loader()
+{
+    VALUE mSqreen = rb_define_module("MiniRacer");
```
SamSaffron, line 120: mSqreen?

SamSaffron:
Overall I like this, is there any impact to load time of the library? 

Additionally are we blocked here on a libv8 change? Is there a PR we need to amend it?

ext/mini_racer_loader/mini_racer_loader.c
lloeki: Whoa 🤦 sorry for this one, fixing it.

lloeki:
> Overall I like this, is there any impact to load time of the library?

With `RTLD_NOW`, a theoretical one, albeit small. I changed it to `RTLD_LAZY` in an attempt to fix Travis CI breaking on Linux, which also spreads it (possibly reducing it overall if some symbols are never used). There's the `$LOAD_PATH` lookup too, quite similar to what an additional `require` does.

I'm looking into the CI break. We've had this working for a while on Linux (including with musl) so I'm not yet sure what the issue is:

```
/home/travis/build/rubyjs/mini_racer/lib/mini_racer.rb:8:in `load': /home/travis/build/rubyjs/mini_racer/lib/mini_racer_extension.so: undefined symbol: _ZTTNSt7__cxx1119basic_istringstreamIcSt11char_traitsIcESaIcEEE (LoadError)
```

> Additionally are we blocked here on a libv8 change?

With the loader, we're not. In the future, if (when?) libv8 manages to be built with `-fvisibility-hidden` (or something to that effect), we could drop the loader.

> Is there a PR we need to amend it?

I'm currently trying to find a simple way to have `libv8_monolith.a` built with local symbols and plan to open a PR there.

lib/mini_racer.rb
```diff
@@ -1,5 +1,12 @@
 require "mini_racer/version"
-require "mini_racer_extension"
+require "mini_racer_loader"
+
+ext_filename = "mini_racer_extension.#{RbConfig::CONFIG['DLEXT']}"
+ext_path = $LOAD_PATH.map { |p| (Pathname.new(p) + ext_filename) }.find { |p| p.file? }
```
SamSaffron, line 5: do we need to walk the whole load path here? can we just rely on __FILE___ or __dir__ here?

lloeki: This would point us to `lib`, or possibly `ext/mini_racer_extension` indeed, but in a hardcoded way. Ruby has this funny way of using one or more of those paths:

```
$ find ~/.gem/ruby/2.7.1 -name 'mini_racer_extension.bundle' | xargs ls -l 
-rwxr-xr-x  1 lloeki  staff  31947216 Sep 30 12:19 /Users/lloeki/.gem/ruby/2.7.1/extensions/x86_64-darwin-19/2.7.0-static/mini_racer-0.3.1/mini_racer_extension.bundle
-rwxr-xr-x  1 lloeki  staff  31947216 Sep 30 12:19 /Users/lloeki/.gem/ruby/2.7.1/gems/mini_racer-0.3.1/ext/mini_racer_extension/mini_racer_extension.bundle
-rwxr-xr-x  1 lloeki  staff  31947216 Sep 30 12:19 /Users/lloeki/.gem/ruby/2.7.1/gems/mini_racer-0.3.1/lib/mini_racer_extension.bundle
```

To reduce the path set down from `$LOAD_PATH`, we could use `Gem::Specification#require_paths` instead:

```
$ bundle exec ruby -rmini_racer -e 'p Gem.loaded_specs["mini_racer"].require_paths'                                                
ruby -rmini_racer -e 'p Gem.loaded_specs["mini_racer"].require_paths'       
["/Users/lloeki/.gem/ruby/2.7.1/extensions/x86_64-darwin-19/2.7.0-static/mini_racer-0.3.1", "lib", "ext"]
```

That'd make the marked line become:

```
Gem.loaded_specs["mini_racer"].require_paths.map { |p| (Pathname.new(p) + ext_filename) }.find { |p| p.file? }
```

WDYT?

lloeki:
Seems like the issue with `_ZTTNSt7__cxx1119basic_istringstreamIcSt11char_traitsIcESaIcEEE` a.k.a `VTT for std::__cxx11::basic_istringstream<char, std::char_traits<char>, std::allocator<char> >` is unrelated to this PR:

- Those two other PR's jobs failed for similar reasons
- I cannot reproduce it on the official `ruby:2.7` image

https://travis-ci.org/github/rubyjs/mini_racer/jobs/714390184

https://travis-ci.org/github/rubyjs/mini_racer/jobs/732159781

lloeki:
I took on myself to add a `Dockerfile` (and `.dockerignore`) to this PR for easy and consistent reproduction with the following command. So as not to block this one I can move it to another PR or altogether remove it though, as you prefer.

```
docker build -t mini_racer . && docker run --rm -it mini_racer bundle exec rake test
docker build --build-arg RUBY_VERSION=2.6 -t mini_racer . && docker run --rm -it mini_racer bundle exec rake test
```

lib/mini_racer.rb
lloeki: I have implemented the change.

SamSaffron: that is much better! thanks.

SamSaffron:
There has been a lot complaints over the past few months regarding mini_racer builds failing on old versions of gcc. I strongly suspect this is all somewhat related.

SamSaffron:
I am fine to merge this, but can you confirm this is going to work on MacOS as well (I can see you allowed for it) 

Another huge one I would love some help on is: 

https://chromium-review.googlesource.com/c/v8/v8/+/2416501

This will be enormously useful (especially for sqreen) cause if we have a switch for single threaded mode you could use it and not need to worry about processes forking anymore and breaking the runtime.

lloeki:
> Another huge one I would love some help on is

Is this tied to this one and that commit?

- https://github.com/rubyjs/mini_racer/issues/175
- https://github.com/rubyjs/mini_racer/commit/86bdc6d9422345dbc2786186b054ec1cbab44bb3

> This will be enormously useful (especially for sqreen) cause if we have a switch for single threaded mode you could use it and not need to worry about processes forking anymore and breaking the runtime.

As it turns out I just happen to have a customer who's using the `parallel` gem (in some yet to be determined way, but it appears to fork so probably not threaded mode) and recently complaining that our V8 isolates crash his app.

I cannot commit on a timeline but this definitely looks worth a shot. I'll bring this up to the team.

lloeki:
> can you confirm this is going to work on MacOS as well (I can see you allowed for it)

Yes, this works on macOS (incidentally I work on macOS first and test on Linux later).

SamSaffron:
Cool, let try this out then! Thanks @lloeki .

PR #180, opened 2020-11-10
Single threaded platform

cataphract:


SamSaffron:
I see, so the big fix we need of master is: 
```
 auto on_fork_for_child = []() {
            exit_lock = PTHREAD_RWLOCK_INITIALIZER;
        };
        pthread_atfork(nullptr, nullptr, on_fork_for_child);

```

Can you make a separate PR for that on master. It is going to take quite a while for V8 to merge single thread platform, and this will give us a tiny bit more safety till then.

ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -205,9 +203,6 @@ static VALUE rb_platform_terminate(VALUE self) {
 }
 
 static void init_v8() {
-    // no need to wait for the lock if already initialized
-    if (current_platform != NULL) return;
```
SamSaffron, line 209: is this really a concern in any practical sense? every time we init a context we call this so this adds 1 lock acquisition to every context creation.

cataphract: This pattern is called double-checked locking, and this particular implementation is unsafe. In amd64 it's safe as long as the compiler doesn't do some reordering of the assignments, but to ensure that you'd need to add compiler fences anyway. You can see a summary and what a correct implementation would like here: https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/

That said, I doubt there is a lot of contention on this lock (which after the initialization is only held to check if the pointer is not null), so I didn't implement the double-checked locking.

cataphract:
> I see, so the big fix we need of master is:
> 
> ```
>  auto on_fork_for_child = []() {
>             exit_lock = PTHREAD_RWLOCK_INITIALIZER;
>         };
>         pthread_atfork(nullptr, nullptr, on_fork_for_child);
> ```
> 
> Can you make a separate PR for that on master. It is going to take quite a while for V8 to merge single thread platform, and this will give us a tiny bit more safety till then.

I've opened https://github.com/rubyjs/mini_racer/pull/181

ext/mini_racer_extension/mini_racer_extension.cc
SamSaffron: fair enough, thanks for explaining!

PR #181, opened 2020-11-11
Fix hang after fork

cataphract:


SamSaffron:
This looks good, going to do another mini_racer release for this!

PR #198, opened 2021-05-06
Prevent segfaults related to infinite marshal recursion

seanmakesgames:
Creates a marshal stackdepth option, allowing protection against exceptional object structures. See testcases for examples.
Been a while since I've made changes in here, would love some critical eyes on the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Gemfile.lock
 lib/mini_racer_extension.so
 lib/mini_racer_loader.so
 *.bundle
+.vscode/
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -115,6 +115,7 @@ typedef struct {
     useconds_t timeout;
     EvalResult* result;
     size_t max_memory;
+    size_t marshal_stackdepth;
 } EvalParams;
 
 typedef struct {
@@ -126,6 +127,7 @@ typedef struct {
     Local<Value> *argv;
     EvalResult result;
     size_t max_memory;
+    size_t marshal_stackdepth;
 } FunctionCall;
 
 enum IsolateFlags {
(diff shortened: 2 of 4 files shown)


seanmakesgames:
After v8 ver upgrade, max data slots seem to have reduced. Need to refactor our usage and allocate some memory to track our runtime state per isolate.

seanmakesgames:
Also, @SamSaffron the locale test you added is failing on my machine on the main branch. Is this expected?
darwin 10.15, ruby 2.6.5

SamSaffron:
@seanmakesgames error is kind of expected until you update libv8-node, try installing latest and updating lockfile etc.

SamSaffron:
I need some time to think about this, can you think of a way of introducing minimal levels of safety here without making this configurable? Maybe we just limit at 1000 depth unconditionally using a const or something like that and the patch can be simpler?

seanmakesgames:
> @seanmakesgames error is kind of expected until you update libv8-node, try installing latest and updating lockfile etc.

Got it. Thanks. I figured the dependency file would have forced the upgrade when I ran `bundle`.

seanmakesgames:
> I need some time to think about this, can you think of a way of introducing minimal levels of safety here without making this configurable? Maybe we just limit at 1000 depth unconditionally using a const or something like that and the patch can be simpler?

I can definitely put that together. For my uses I need 100. :)
I'll put together a few different PRs.

seanmakesgames:
> > > @seanmakesgames error is kind of expected until you update libv8-node, try installing latest and updating lockfile etc.
> > 
> > 
> > Got it. Thanks. I figured the dependency file would have forced the upgrade when I ran `bundle`.
> 
> I deleted my lock file and rebundled and am still having the error. What version of libv8-node am I looking for here?
> [Gemfile.lock.txt](https://github.com/rubyjs/mini_racer/files/6442755/Gemfile.lock.txt)

This was fixed by #200

seanmakesgames:
> I need some time to think about this, can you think of a way of introducing minimal levels of safety here without making this configurable? Maybe we just limit at 1000 depth unconditionally using a const or something like that and the patch can be simpler?

@SamSaffron I just tried to implement this with a fixed depth -- but I can't without #201 because we are at data slot max already: 4/4. 

I need flags like MEM_SOFTLIMIT_VALUE and MEM_SOFTLIMIT_REACHED to be able to track the stack depth and report it back safely across thread contexts and such. I'll push my WIP -

seanmakesgames:
I put the non-configurable stack max into #202 -- needs 201 to work. Happy to update them however.

seanmakesgames:
implemented by #202

PR #201, opened 2021-05-08
Refactor isolate data usage into IsolateData

seanmakesgames:
Pack the data into smaller usage; reinterpreting the voidptrs as big uints and not using any pointer indirection.

Removes a bunch of static casts, and makes room for more fields / flags.

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -128,11 +128,56 @@ typedef struct {
     size_t max_memory;
 } FunctionCall;
 
-enum IsolateFlags {
-    IN_GVL,
-    DO_TERMINATE,
-    MEM_SOFTLIMIT_VALUE,
-    MEM_SOFTLIMIT_REACHED,
+#define SLOT_BOUNDARY 1000 // every thousand denotes slot boundary
+
+class IsolateData {
+public:
+    enum Flag {
+        // first flags are bitfield
+        //  max count: sizeof(uintptr_t) * 8
+        IN_GVL, // whether we are inside of ruby gvl or not
+        DO_TERMINATE, // terminate as soon as possible
+        MEM_SOFTLIMIT_REACHED, // we've hit the memory soft limit
+
+        // softlimit size is full-sized uintptr_t
+        MEM_SOFTLIMIT_VALUE = 1 * SLOT_BOUNDARY,
+    };
+
+    static uintptr_t Get(Isolate *isolate, Flag flag) {
+        uintptr_t slotData, flagMask, retval;
+        uint slot = flag / SLOT_BOUNDARY;
+        slotData = reinterpret_cast<uintptr_t>(isolate->GetData(slot));
+
+        switch (slot) {
+        case 0:
+            flagMask = 1 << flag;
+            retval = (slotData & flagMask) == flagMask;
+            return retval;
+        case 1:
+            return slotData;
+        }
+
+        throw std::runtime_error("MINI_RACER_DEV_ERROR Flags didn't match switch statement in Get.");
+    }
+
+    static void Set(Isolate *isolate, Flag flag, uintptr_t value) {
+        uintptr_t slotData, flagMask, setval;
+        uint slot = flag / SLOT_BOUNDARY;
+
+        switch (slot) {
+        case 0:
+            slotData = reinterpret_cast<uintptr_t>(isolate->GetData(slot));
+            flagMask = 1 << flag;
+            setval = !!value ? slotData | flagMask : slotData & ~flagMask;
+            isolate->SetData(slot, reinterpret_cast<void*>(setval));
+            return;
+        case 1:
+            isolate->SetData(slot, reinterpret_cast<void*>(value));
+            return;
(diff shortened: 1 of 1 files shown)


ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -128,11 +128,56 @@ typedef struct {
...
+        switch (slot) {
+        case 0:
+            flagMask = 1 << flag;
+            retval = (slotData & flagMask) == flagMask;
+            return retval;
+        case 1:
+            return slotData;
+        }
+
+        throw std::runtime_error("MINI_RACER_DEV_ERROR Flags didn't match switch statement in Get.");
```
SamSaffron, line 160: this is going to be the first instance of `throw` in the script, is it safe? should simply return "not set" for unknown slot?

SamSaffron:
This is great, I support this change, gives us more flexibility.

ext/mini_racer_extension/mini_racer_extension.cc
seanmakesgames: Really it's a dev-time error. I'd love for it to be a compile time error somehow -- there's no return value here that makes sense. (it's a uint, so -1 and 0 are both values to return that would be real bad and cause hard-to-debug issues)
We could assert (which is basically an exception as well) -- throwing is not 'safe' but should never be hit unless there's a bug in our implementation. 

I'm 100% open to ideas here. This was just the best I could come up with in the time I thought about it. :)

SamSaffron: as long as it is safe I guess I am fine with it, I guess it will cause a segfault if we mess this up?

seanmakesgames:
> Curious, what about a different packing mechanism?
> 
> We use 2 bits to denote the data type, and then follow with the data? We would have to walk all the "sizes" first which is not super efficient, but maybe it is cleaner?
> 
> 00 = bit
> 01 = uint
> 10 = ulong
> 11 = ulong long

I would have packed this a lot tighter, but I have no idea if the data pointers are contiguous.

Also, since we are hitting these things during marshalling, we want to make sure that the compiler will optimize to something very fast and simple. 

If the data void*s are actually a contiguous block of memory, I would just use masks against them starting at the beginning. That's super fast and would be very simple code. Reason this is more complicated now is I'm being careful of the boundaries.

SamSaffron:
> If the data void*s are actually a contiguous block of memory 

Reading through v8 it appears that it all boils down to `std::atomic_store_explicit` this appears to be happy with storing stuff up to 64 bits. Pointer storage seems optional which may be a concern? 

https://en.cppreference.com/w/cpp/atomic/atomic

Can we just smash all the data we need into 64 bits? Happy to drop a tiny bit of fidelity on memory limits.

seanmakesgames:
> > If the data void*s are actually a contiguous block of memory
> 
> Reading through v8 it appears that it all boils down to `std::atomic_store_explicit` this appears to be happy with storing stuff up to 64 bits. Pointer storage seems optional which may be a concern?
> 
> https://en.cppreference.com/w/cpp/atomic/atomic
> 
> Can we just smash all the data we need into 64 bits? Happy to drop a tiny bit of fidelity on memory limits.

Yeah, we should be able to. I need 36 bits without the memory limit, which we can just do 24bits of precision or something. 

I've got a new plan that simplifies things. Will post a new one when I've got it.

seanmakesgames:
Switched to implementation using union and bitfield struct. (removes the bit manipulation portion of the packing, and the maintenance of it.)

Having a 'covered switch statement' allows the compiler to (sadly just warn in this case) detect and alert on having a flag that's not covered:

```
../../../../ext/mini_racer_extension/mini_racer_extension.cc:145:17: warning: enumeration value 'dummy' not handled in switch [-Wswitch]
        switch (flag) {
                ^
../../../../ext/mini_racer_extension/mini_racer_extension.cc:145:17: note: add missing switch cases
        switch (flag) {
                ^
../../../../ext/mini_racer_extension/mini_racer_extension.cc:151:5: warning: non-void function does not return a value in all control paths [-Wreturn-type]
    }
    ^
../../../../ext/mini_racer_extension/mini_racer_extension.cc:155:17: warning: enumeration value 'dummy' not handled in switch [-Wswitch]
        switch (flag) {
                ^
../../../../ext/mini_racer_extension/mini_racer_extension.cc:155:17: note: add missing switch cases
        switch (flag) {
                ^
```

Additional comments about the implementation and static_assert in code. Currently this forces mini_racer to not work in 32bit systems-- I don't know if that was already the case for some reason. We can adapt the fix to support <64bit systems if necessary.

ext/mini_racer_extension/mini_racer_extension.cc
seanmakesgames: Gone in latest impl.

seanmakesgames:
Which warning?

The one I mention in my comment is a compile time safety net when the count
of the flag enums don’t match our switch statement. Otherwise don’t know
which warning you mean.

seanmakesgames:
Yep! Thanks-
Got a bit of refactoring to do, but should be ready soon.

PR #202, opened 2021-05-10
Configurable max marshal stack depth

seanmakesgames:
builds on #201 (bit of a weird git history at the moment) -- will sort itself out with some squashing and merging other PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ Gemfile.lock
 lib/mini_racer_extension.so
 lib/mini_racer_loader.so
 *.bundle
+.vscode/
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ It is possible to set V8 Runtime flags:
 MiniRacer::Platform.set_flags! :noconcurrent_recompilation, max_inlining_levels: 10
 ```
 
-This can come in handy if you want to use MiniRacer with Unicorn, which doesn't seem to alwatys appreciate V8's liberal use of threading:
+This can come in handy if you want to use MiniRacer with Unicorn, which doesn't seem to always appreciate V8's liberal use of threading:
 ```ruby
 MiniRacer::Platform.set_flags! :noconcurrent_recompilation, :noconcurrent_sweeping
 ```
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ context.eval 'var a = new Array(10000); while(true) {a = a.concat(new Array(1000
 
 ### Object marshal max stackdepth support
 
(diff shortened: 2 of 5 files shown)


SamSaffron:
Any chance you can measure the impact here? I am guessing it is negligible considering huge costs of cross vm serialization which would dwarf this. 

As long as you confirm this is totally negligible in the big scheme, perfectly happy to merge and not bother making this configurable.

Note there are going to still probably be hacks that can trigger explosions, like an object with 100k leaves each of which has another 100k leaves. Even at 5 deep people may be able to segfault? Do we just rely on memory limits for these cases?

seanmakesgames:
For my use I would prefer a stack depth of 100 (mongodb max object size), and have it configureable.

Most of the resource-based DoSes have been mitigated by stuff that has already been done like memory caps. There are probably a few of these left, but much more edge-case-y.

I'm going to just add back in the configurability to this PR. It's a bit more code, but not that much. I'll also do a run of the benchmarking code if I can figure it out. Let me know if you have any comments about the implementation itself thus far.

seanmakesgames:
You're right. Impact on perf with it on is noise:

With it set at 512:
```
Benching with mini_racer
-----on-------
mini_racer minify discourse_app.js 9613.315ms
-----on-------
mini_racer minify discourse_app_minified.js 8677.78ms
-----on-------
mini_racer minify discourse_app.js twice (2 threads) 10136.720000000001ms
```

With it switched off:
```
Benching with mini_racer
mini_racer minify discourse_app.js 9512.943ms
mini_racer minify discourse_app_minified.js 8620.75ms
mini_racer minify discourse_app.js twice (2 threads) 10319.359999999999ms
```

With the code commented out:
```
Benching with mini_racer
mini_racer minify discourse_app.js 9581.952ms
mini_racer minify discourse_app_minified.js 8766.073ms
mini_racer minify discourse_app.js twice (2 threads) 10500.054ms
```

SamSaffron:
Oh I am not sure that is the right micro bench here :thinking: 

I would go for say something like this?  


```
1000.times do
   ctx.eval({a: {b: {c: "d"}}})
end
```

Can you rerun numbers? 

Happy for the default to be 1023.

SamSaffron:
OK, since this is arguably measurable lets go with default off. 

Can you amend the PR to also explain about this knob in the README?

seanmakesgames:
Readme updates should be in there. I'll update the default to off.

seanmakesgames:
Default is off again. README was updated.

Did you want me to edit the PR description to talk more about the feature that was added here, or is the readme modification sufficient?

SamSaffron:
I think this is good as is, will think over the next few days if it makes sense to have the safer default, given this will only impact outlier code.

PR #205, opened 2021-06-03
gemspec: Drop configuration for executables

olleolleolle:
This gem exposes 0 executables.


SamSaffron:
very true, nice catch, thanks!

PR #207, opened 2021-06-04
Travis CI: Attempt to use rvm-installed Bundler, RubyGems

olleolleolle:
This is an attempt to make the CI configuration smaller, more confident. I am guessing that we are not relying on Bundler 1.x-specific behavior, so this commit is a try-out of that.

I was inspired to do this when I saw that the `gem update --system` change took very long in builds.

bild

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,4 @@ jobs:
       os: osx
       osx_image: xcode12.2
 dist: xenial
-before_install:
-  - gem update --system
-  - gem install bundler -v 1.16.2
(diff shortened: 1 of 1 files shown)


SamSaffron:
Sure I am happy to give this a shot.

olleolleolle:
There were failed builds, but I doubt they were about the RubyGems or Bundler:

Arm: 2.5, 2.7 had "stack smashing detected" in them. 2.6 had a "malloc reported memory corruption"

PR #212, opened 2021-07-06
False positive on OOM error

seanmakesgames:
It took us a few weeks to nail down this repro because it was so incredibly finicky.

I recently switched from an execution context which ran one eval with all the code into it, into multiple evals on the same context. Because of this, we ran into this crucial and very hard to isolate issue where the GC callback was happening between `IsolateData::Init` and `IsolateData::Set` -- when the GC callback was hit, the 'max memory' field was 0 meaning the check of `current memory > max memory` would always trigger.

In this incredibly hard to reproduce (we were never able to create a mini_racer cross-machine portable repro) case, it would erroneously report an out of memory.

The fix for this is to move the setter and callback right next to the data init call.

We've run this change on the production servers and verified all the live repro cases we had previously and by the time you take a look at this, it will have baked for a while.

Please let me know if you have any questions.

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -469,7 +469,6 @@ nogvl_context_eval(void* arg) {
 
     if (eval_params->max_memory > 0) {
         IsolateData::Set(isolate, IsolateData::MEM_SOFTLIMIT_MAX, eval_params->max_memory);
-printf("lim: %lu %lu aaa\n", (size_t)IsolateData::Get(isolate, IsolateData::MEM_SOFTLIMIT_MAX), eval_params->max_memory);
         if (!isolate_info->added_gc_cb) {
         isolate->AddGCEpilogueCallback(gc_callback);
             isolate_info->added_gc_cb = true;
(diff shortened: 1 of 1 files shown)


ext/mini_racer_extension/mini_racer_extension.cc
```diff
@@ -467,6 +467,14 @@ nogvl_context_eval(void* arg) {
 
     IsolateData::Init(isolate);
 
+    if (eval_params->max_memory > 0) {
+        IsolateData::Set(isolate, IsolateData::MEM_SOFTLIMIT_MAX, eval_params->max_memory);
+        if (!isolate_info->added_gc_cb) {
+        isolate->AddGCEpilogueCallback(gc_callback);
```
nightpool, line 473: indentation

nightpool (commented):
wow, great find! I don't know enough about the C code here to give you a thorough review, but it looks like you missed some indentation changes:

ext/mini_racer_extension/mini_racer_extension.cc
nightpool:
```suggestion
            isolate->AddGCEpilogueCallback(gc_callback);
```

seanmakesgames:
> wow, great find! I don't know enough about the C code here to give you a thorough review, but I did notice one small issue:

Thank goodness. That one has annoyed me for a long time. Somehow after all the stuff, I missed it in the edit. :D

SamSaffron:
looks like a good catch to me! thanks.

Issue #190, opened 2021-04-10
0.4.0 crashes on aarch64-linux for Ruby 2.7 + 3.0; LoadError with 2.5 and 2.6 (via Docker for Mac on M1)


tisba:
I used the following to check if `mini_racer` works with Docker for Mac on Apple Silicon. `RUBY_PLATFORM` is `aarch64-linux`.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "mini_racer", "0.4.0"
  gem "libv8-node", "15.14.0.1"
end

puts "RUBY_VERSION : #{RUBY_VERSION}"
puts "RUBY_PLATFORM: #{RUBY_PLATFORM}"
puts "MiniRacer::LIBV8_NODE_VERSION: #{MiniRacer::LIBV8_NODE_VERSION}"
puts "Libv8::Node::VERSION: #{Libv8::Node::VERSION}"
puts "Libv8::Node::NODE_VERSION: #{Libv8::Node::NODE_VERSION}"
puts "Libv8::Node::LIBV8_VERSION: #{Libv8::Node::LIBV8_VERSION}"

ctx = MiniRacer::Context.new
ctx.eval("1+1")
```

I'm on [Docker for Mac 3.3.3 (64133)](https://docs.docker.com/docker-for-mac/release-notes/#docker-desktop-333). For testing, I run the above script like this:

```
docker run -it --rm -v "$(pwd)":/app ruby:2.7.2 ruby /app/minimal.rb
```

**Ruby 3** ❌
* Ruby 3.0.0 and 3.0.1 seems to work fine for the example from above ✅ 
* …but seems to have the same issue as with Ruby 2.7, see https://github.com/rubyjs/mini_racer/issues/190#issuecomment-841808260 ❌

**Ruby 2.7** ❌

Ruby 2.7.0, 2.7.1, 2.7.2 and 2.7.3 results in `*** stack smashing detected ***: <unknown> terminated`. Not sure how to supply more detail here.

**Ruby 2.6 and 2.5** ❌

Ruby 2.5.8 and 2.6.7 gives a `LoadError`:

```
/usr/local/bundle/gems/libv8-node-15.12.0.0.beta1-aarch64-linux/lib/libv8/node.rb:2:in `require': cannot load such file -- libv8-node/location (LoadError)
[...]

PR #214, opened 2021-07-09
bump required ruby version to >= 2.6

tisba:
In https://github.com/rubyjs/mini_racer/commit/06cca5bf190017516650e2f4a1d27c586c12dfda#diff-30d3134eb72fcd96ce343d5c0359684a5545843b47d7cba45732d3e360cf322f the required ruby version was bumped to >= 2.5.

I just stumbled across [Ruby Maintenance Branches](https://www.ruby-lang.org/en/downloads/branches/) where it lists Ruby 2.5's EOL date to 2021-03-31. Based on [@SamSaffron's comment](https://github.com/rubyjs/mini_racer/issues/190#issuecomment-824609197) and the README stating that only non-EOL'd versions of Ruby will be supported, I guess, we could bump the version once more?

It's just a suggestion, I just stumbled across the Ruby maintenance overview.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,6 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-          - '2.5'
           - '2.6'
           - '2.7'
           - '3.0'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 os: linux
 rvm:
-  - 2.5
   - 2.6
   - 2.7
   - 3.0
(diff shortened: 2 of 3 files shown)


SamSaffron:
This is a bit tricky, I very much prefer all my gems only ever support non EOL rubies (people wanting EOL rubies can use old gems). @lloeki wanted a less aggressive target.

@lloeki are you OK with me merging this.

lloeki:
@SamSaffron I'm fine with it.

I don't rely on mini_racer directly yet and still have some stuff to upstream before I can do that, so there's no reason to change the current policy right now.

PR #237, opened 2022-02-10
CI: Run gem update --system on Linux builds

tisba:
This is an attempt to fix the currently broken builds on linux/amd64.

I'm not sure if it's a good idea to "blindly" update rubygems to the latest version, but I thought it's worth a try if it fixes the builds.

https://github.com/rubyjs/mini_racer/runs/4892980034?check_suite_focus=true for example is failing with, which looked pretty obvious to me: the incorrect libv8-node version/architecture has been installed:

```
[…]
linking shared-object mini_racer_extension.so
g++: error: /usr/local/bundle/gems/libv8-node-16.10.0.0-x86_64-linux/vendor/v8/x86_64-linux/libv8/obj/libv8_monolith.a: No such file or directory
gmake: *** [Makefile:262: mini_racer_extension.so] Error 1
rake aborted!
[…]
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,8 @@ jobs:
         run: docker exec -w "${PWD}" ${{ steps.container.outputs.id }} apk add --no-cache build-base linux-headers bash python2 python3 git curl tar clang binutils-gold
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Update Rubygems
+        run: docker exec -w "${PWD}" ${{ steps.container.outputs.id }} gem update --system
       - name: Bundle
         run: docker exec -w "${PWD}" ${{ steps.container.outputs.id }} bundle install
       - name: Compile


lloeki:
I was wondering about that too. One reason I kept it at the default using the official docker hub images was to make sure the experience is ok with these rubygems too, and/or be able to provide workarounds to folks that can't update their rubygems/bundler for whatever reason.

tisba:
But we do know, that the default Rubygem versions are broken.

IMO it is better to have CI working. From what I understand from the current issues with Rubygems there is not really an alternative.

tisba:
Can you trigger the build to run, @lloeki? I'm not 100% sure it actually fixes the issue because I'm having issues reproducing the failure locally 🤔

tisba:
This was finally done via https://github.com/rubyjs/mini_racer/pull/256! Thx @gschlager for getting this merged 🥳

PR #238, opened 2022-02-13
Remove travis-ci.org

tisba:
> Since June 15th, 2021, the building on [travis-ci.org](http://www.travis-ci.org/) is ceased.

I updated the status badge to point to Github Actions.

Since probably nobody is using Travis anymore and the info section is the REAME seemed very outdated to me, I also removed the reference to Travis there as well.


SamSaffron:
oops conflicting now, I merged in the wrong order, can you rebase?

PR #239, opened 2022-02-13
Readme: Add note about V8 single threaded platform mode

tisba:
I'm building up a little troubleshooting guide and noticed that we do not mention the support of the single threaded platform mode anywhere.

Also: I was looking when Ruby 2.6 is going EOL'd (it's 2022-03-31) and changed the link to the correct  "Ruby Maintenance Branches" site where there is an overview.


SamSaffron:
Very handy to have this in the readme!

PR #253, opened 2022-07-12
Implement MiniRacer for TruffleRuby

bjfish:
This PR implements mini_racer for TruffleRuby using the GraalVM's JavaScript implementation.

It can be run locally using:
```
% rbenv install truffleruby+graalvm-dev
% rbenv shell truffleruby+graalvm-dev
% gu install js
% export TRUFFLERUBYOPT='--jvm --polyglot'
% bundle install 
% bundle exec rake compile
% bundle exec rake test
```

A previous PR with some code review is here: https://github.com/bjfish/mini_racer/pull/1

cc: @eregon

### DIFF
--- a/lib/mini_racer/truffleruby.rb
+++ b/lib/mini_racer/truffleruby.rb
@@ -1,0 +1,353 @@
+# frozen_string_literal: true
+
+module MiniRacer
+
+  class Context
+
+    class ExternalFunction
+      private
+
+      def notify_v8
+        name = @name.encode(::Encoding::UTF_8)
+        wrapped = lambda do |*args|
+          converted = @parent.send(:convert_js_to_ruby, args)
+          begin
+            result = @callback.call(*converted)
+          rescue Polyglot::ForeignException => e
+            e = RuntimeError.new(e.message)
+            e.set_backtrace(e.backtrace)
+            @parent.instance_variable_set(:@current_exception, e)
+            raise e
+          rescue => e
+            @parent.instance_variable_set(:@current_exception, e)
+            raise e
+          end
+          @parent.send(:convert_ruby_to_js, result)
+        end
+
+        if @parent_object.nil?
+          # set global name to proc
+          result = @parent.eval_in_context('this')
+          result[name] = wrapped
+        else
+          parent_object_eval = @parent_object_eval.encode(::Encoding::UTF_8)
+          begin
+            result = @parent.eval_in_context(parent_object_eval)
+          rescue Polyglot::ForeignException, StandardError => e
+            raise ParseError, "Was expecting #{@parent_object} to be an object", e.backtrace
+          end
+          result[name] = wrapped
+          # set evaluated object results name to proc
+        end
+      end
+    end
+
+    def heap_stats
+      {
+        total_physical_size: 0,
+        total_heap_size_executable: 0,
+        total_heap_size: 0,
+        used_heap_size: 0,
+        heap_size_limit: 0,
+      }
+    end
+
+    def stop
+      if @entered
+        @context.stop
+        @stopped = true
+        stop_attached
+      end
+    end
+
+    private
+
+    @context_initialized = false
+    @use_strict = false
+
+    def init_unsafe(isolate, snapshot)
+      unless defined?(Polyglot::InnerContext)
+        raise "TruffleRuby #{RUBY_ENGINE_VERSION} does not have support for inner contexts, use a more recent version"
+      end
+
+      unless Polyglot.languages.include? "js"
+        warn "You also need to install the 'js' component with 'gu install js' on GraalVM 22.2+", uplevel: 0 if $VERBOSE
+      end
+
+      @context = Polyglot::InnerContext.new(on_cancelled: -> { 
+        raise ScriptTerminatedError, 'JavaScript was terminated (either by timeout or explicitly)' 
+      })
+      Context.instance_variable_set(:@context_initialized, true)
+      @js_object = @context.eval('js', 'Object')
+      @isolate_mutex = Mutex.new
+      @stopped = false
+      @entered = false
+      @has_entered = false
+      @current_exception = nil
+      if isolate && snapshot
+        isolate.instance_variable_set(:@snapshot, snapshot)
+      end
+      if snapshot
+        @snapshot = snapshot
+      elsif isolate
+        @snapshot = isolate.instance_variable_get(:@snapshot)
+      else
+        @snapshot = nil
+      end
+      @is_object_or_array_func, @is_time_func, @js_date_to_time_func, @is_symbol_func, @js_symbol_to_symbol_func, @js_new_date_func, @js_new_array_func = eval_in_context <<-CODE
+        [
+          (x) => { return (x instanceof Object || x instanceof Array) && !(x instanceof Date) && !(x instanceof Function) },
+          (x) => { return x instanceof Date },
+          (x) => { return x.getTime(x) },
+          (x) => { return typeof x === 'symbol' },
+          (x) => { var r = x.description; return r === undefined ? 'undefined' : r },
+          (x) => { return new Date(x) },
+          (x) => { return new Array(x) },
+        ]
+      CODE
+    end
+
+    def dispose_unsafe
+      @context.close
+    end
+
+    def eval_unsafe(str, filename)
+      @entered = true
+      if !@has_entered && @snapshot
+        snapshot_src = encode(@snapshot.instance_variable_get(:@source))
+        begin
+          eval_in_context(snapshot_src, filename)
+        rescue Polyglot::ForeignException => e
+          raise RuntimeError, e.message, e.backtrace
+        end
+      end
+      @has_entered = true
+      raise RuntimeError, "TruffleRuby does not support eval after stop" if @stopped
+      raise TypeError, "wrong type argument #{str.class} (should be a string)" unless str.is_a?(String)
+      raise TypeError, "wrong type argument #{filename.class} (should be a string)" unless filename.nil? || filename.is_a?(String)
+
+      str = encode(str)
+      begin
+        translate do
+          eval_in_context(str, filename)
+        end
+      rescue Polyglot::ForeignException => e
+        raise RuntimeError, e.message, e.backtrace
+      rescue ::RuntimeError => e
+        if @current_exception
+          e = @current_exception
+          @current_exception = nil
+          raise e
+        else
+          raise e, e.message
+        end
+      end
+    ensure
+      @entered = false
+    end
+
+    def call_unsafe(function_name, *arguments)
+      @entered = true
+      if !@has_entered && @snapshot
+        src = encode(@snapshot.instance_variable_get(:source))
+        begin
+          eval_in_context(src)
+        rescue Polyglot::ForeignException => e
+          raise RuntimeError, e.message, e.backtrace
+        end
+      end
+      @has_entered = true
+      raise RuntimeError, "TruffleRuby does not support call after stop" if @stopped
+      begin
+        translate do
+          function = eval_in_context(function_name)
+          function.call(*convert_ruby_to_js(arguments))
+        end
+      rescue Polyglot::ForeignException => e
+        raise RuntimeError, e.message, e.backtrace
+      end
+    ensure
+      @entered = false
+    end
+
+    def create_isolate_value
+      # Returning a dummy object since TruffleRuby does not have a 1-1 concept with isolate.
+      # However, code and ASTs are shared between contexts.
+      Isolate.new
+    end
+
+    def isolate_mutex
+      @isolate_mutex
+    end
+
+    def translate
+      convert_js_to_ruby yield
+    rescue Object => e
+      message = e.message
+      if @current_exception
+        raise @current_exception
+      elsif e.message && e.message.start_with?('SyntaxError:')
+        error_class = MiniRacer::ParseError
+      elsif e.is_a?(MiniRacer::ScriptTerminatedError)
+        error_class = MiniRacer::ScriptTerminatedError
+      else
+        error_class = MiniRacer::RuntimeError
+      end
+
+      if error_class == MiniRacer::RuntimeError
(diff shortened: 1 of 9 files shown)


eregon:
Some extra details: this is a pure-Ruby implementation of the gem using the GraalVM Polyglot API and inner contexts for isolation.
It uses the same interface as the C extension and reuses all the Ruby code in `lib/`.
It does not use `libv8-node` at all.

This fixes https://github.com/oracle/truffleruby/issues/1827.

SamSaffron:
Wow, Amazing, thank you so much

SamSaffron:
I hope to get this reviewed early next week, @gschlager is super interested in fully testing this and he is back next week.

lib/mini_racer/truffleruby.rb
```diff
@@ -0,0 +1,355 @@
...
+      @isolate_mutex
+    end
+
+    def translate
+      convert_js_to_ruby yield
+    rescue Object => e
+      message = e.message
+      if @current_exception
+        raise @current_exception
+      elsif e.message.start_with?('SyntaxError:')
```
gschlager, line 189: I'm seeing a bunch of errors in this line when I'm running specs from Discourse that are specific to `MiniRacer`.

```
bundle exec rspec spec/lib/pretty_text_spec.rb
```

Here's the error:
```
NoMethodError:
undefined method `start_with?' for nil:NilClass
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-33ab13dc022e/lib/mini_racer/truffleruby.rb:189:in `translate'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-33ab13dc022e/lib/mini_racer/truffleruby.rb:131:in `eval_unsafe'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-33ab13dc022e/lib/mini_racer.rb:228:in `block (2 levels) in eval'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-33ab13dc022e/lib/mini_racer.rb:369:in `timeout'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-33ab13dc022e/lib/mini_racer.rb:227:in `block in eval'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-33ab13dc022e/lib/mini_racer.rb:225:in `synchronize'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-33ab13dc022e/lib/mini_racer.rb:225:in `eval'
# ./lib/pretty_text.rb:245:in `block in markdown'
# ./lib/pretty_text.rb:605:in `block in protect'
# ./lib/pretty_text.rb:604:in `synchronize'
# ./lib/pretty_text.rb:604:in `protect'
# ./lib/pretty_text.rb:188:in `markdown'
# ./lib/pretty_text.rb:317:in `cook'
# ./spec/lib/pretty_text_spec.rb:1657:in `block (3 levels) in <top (required)>'
# ./spec/rails_helper.rb:278:in `block (2 levels) in <top (required)>'
# /home/gerhard/.gem/truffleruby/3.0.3/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# ------------------
# --- Caused by: ---
# Polyglot::ForeignException:
```

eregon: I think that's most likely caused by passing a Ruby String to JS, and JS not yet handling that correctly (calling Ruby methods on it vs JS String prototype functions).

@bjfish You need to handle String like this in `convert_ruby_to_js`:
```ruby
        when String
          Truffle::Interop.as_truffle_string value
```
from https://github.com/bjfish/mini_racer/pull/1#discussion_r916804519 and https://github.com/rails/execjs/pull/115/files

Also add the `if e.message && e.message.start_with?('SyntaxError:')` on this line to help debug.

eregon: The handling of String is already there, but String needs to be removed from `when nil, true, false, Integer, Float, String`

lib/mini_racer/truffleruby.rb
```diff
@@ -0,0 +1,353 @@
...
+
+    class ExternalFunction
+      private
+
+      def notify_v8
+        name = @name.encode(::Encoding::UTF_8)
+        wrapped = lambda do |*args|
+          converted = @parent.send(:convert_js_to_ruby, args)
+          begin
+            result = @callback.call(*converted)
```
gschlager, line 15: MiniRacer with V8 seems to call the `@callback` with unfrozen strings while TruffleRuby uses frozen strings. I don't see a test for it, but I think we should keep that behavior. Would it be possible to have unfrozen strings in `converted`?

Here's the spec error I'm seeing in Discourse:
```
FrozenError:
can't modify frozen String

# <internal:core> core/string.rb:882:in `gsub!'
# ./lib/pretty_text/helpers.rb:112:in `category_tag_hashtag_lookup'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer.rb:273:in `block in attach'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer/truffleruby.rb:15:in `block in notify_v8'
# (mini_racer):13:in `addHashtag'
# (mini_racer):93:in `applyRule'
# (mini_racer):154:in `textPostProcess'
# (mini_racer):173:in `replacer'
# (mini_racer):69:in `textReplace'
# (mini_racer):177:in `:anonymous'
# (mini_racer):4221:in `Core.process'
# (mini_racer):8431:in `MarkdownIt.parse'
# (mini_racer):8446:in `MarkdownIt.render'
# (mini_racer):578:in `cook'
# (mini_racer):126:in `cook'
# (mini_racer):1:in `:program'
# <internal:core> core/truffle/polyglot.rb:37:in `eval'
# (mini_racer):1:in `eval_in_context'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer/truffleruby.rb:132:in `block in eval_unsafe'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer/truffleruby.rb:184:in `translate'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer/truffleruby.rb:131:in `eval_unsafe'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer.rb:228:in `block (2 levels) in eval'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer.rb:369:in `timeout'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer.rb:227:in `block in eval'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer.rb:225:in `synchronize'
# /home/gerhard/.gem/truffleruby/3.0.3/bundler/gems/mini_racer-6206a1e15d3c/lib/mini_racer.rb:225:in `eval'
# ./lib/pretty_text.rb:245:in `block in markdown'
# ./lib/pretty_text.rb:605:in `block in protect'
# ./lib/pretty_text.rb:604:in `synchronize'
# ./lib/pretty_text.rb:604:in `protect'
# ./lib/pretty_text.rb:188:in `markdown'
# ./lib/pretty_text.rb:317:in `cook'
# ./spec/lib/pretty_text_spec.rb:1421:in `block (2 levels) in <top (required)>'
# ./spec/rails_helper.rb:278:in `block (2 levels) in <top (required)>'
# /home/gerhard/.gem/truffleruby/3.0.3/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# <internal:core> core/kernel.rb:376:in `load'
# ------------------
# --- Caused by: ---
# Polyglot::ForeignException:
```

eregon: By design for general GraalVM interoperability, Strings from other languages are considered immutable/frozen.
So JS Strings are immutable, and when we get a corresponding Ruby String for it, it is frozen on purpose to reflect the fact that mutating the Ruby String wouldn't mutate the JS String (not possible).

I think it would be better to modify the caller code to use `.dup` or `.gsub` instead, rather than relying on a library (or a gem) returning a mutable string, that always seems an anti-pattern to me, to assume that a library returns a mutable string, and mutate it (e.g., what if the library uses that string in a Hash/cache, then you'd break the Hash/cache; or if the library would like to return a constant string, it would need to be frozen to be safe).

That said, the semantics of mini_racer for passing arguments seems to be more or less "the same as JSON.load(JSON.dump(obj))", and so a "copy" is expected.
If needed, we can change this by calling `.dup` in `convert_js_to_ruby` for strings, but I wonder if the current behavior is not safer for users. IMHO it encourages the better practice to not assume that the method returns mutable strings and you can mutate it without other side effects in the system.

gschlager: I was quite surprised that MiniRacer (V8) passes mutable strings to callback functions and returns mutable strings when you call `.eval`. I agree, it would be best to always use frozen strings and let the caller decide what to do with it. I'm in favor of introducing a breaking change that makes those methods work with frozen strings even for V8.

@SamSaffron What's your opinion on this subject? In either case, I think there should be tests that ensure/document that behavior.

SamSaffron: Oh I see the dilema here. 

In the V8 embedding at the moment we always do copies on the boundary, I can get that certain specific callers may get annoyed if they have to make yet another just to enforce this contract.

In the case of Discourse this is actually quite wasteful, given we would be copying potentially a 100k long string twice just so we can mutate it.

Maybe a to keep existing performance of MRI we add a tiny line to Discourse here? 

````
# truffle mini racer compatability 
string = string.dup if string.frozen?
```

Alternatives. I can think about

1. The contract is to provide mutable strings on boundary
    - backwards compat
    - most efficient for v8 embedding
    - wasteful in truffle land due to unconditional `.dup`
2. "Flexible" contract
    - callers need to remember to do more work in truffle case
3. Change contract to require frozen strings
    - backwards incompatible
    - any mutating code will need an extra copy of a potentially huge string
    - consistent with truffle implementation
 4. Hybrid, flag based approach
    - If you need a mutable string you do your eval with a special flag
    - potentially at the global level you define if you always want mutable back or not

I am tempted to just go with (1) here and add specification for it. Cause `.dup` is presumably very cheap in truffle and only starts getting expensive once you mutate?

eregon: I think either 1 or 3 are best.
I like 3 because of the safe semantics, and IMHO a method mutating a string argument is almost always a bug.
But I understand compatibility is important of course, and that's why I think it should be both are frozen or both are not.

Yes, performance-wise the `.dup` should be pretty cheap on TruffleRuby once JIT-compiled, probably free even (the allocation of the RubyString before the `.dup` will be escape-analyzed away).
`.dup` is also cheap on CRuby since it's copy-on-write.

Mutating a string OTOH is never free, especially on CRuby where `gsub!` will e.g. have to copy all bytes after each match/replacement (on TruffleRuby Ropes and notably lazy substrings are used so there are no byte copies going on for `gsub!`).

eregon: I think in this PR it doesn't make much sense to change semantics, and being compatible with the current behavior is more important. @bjfish Could you add that `.dup`?

bjfish: @eregon Yes, I've added `.dup` to the `String` branch in `convert_js_to_ruby`.

@gschlager Please retry this spec.

gschlager: @bjfish Thank you! It's looking good. All specs for `PrettyText` in Discourse are green. I can't run all Discourse specs with TruffleRuby, because I'm running out of memory (32GB), but `PrettyText` has the most important tests.

@SamSaffron I think this is good to go. :+1:

gschlager (approved):

lib/mini_racer/truffleruby.rb
eregon: > I can't run all Discourse specs with TruffleRuby, because I'm running out of memory (32GB)

We should look into that at some point, maybe something leaks. Which commands did you use to run the specs or is there documentation we could follow to run Discourse specs?

SamSaffron:
Amazing work, thanks heaps.

@lloeki plans to update our v8 version, when we do that I think we can announce a new release of mini_racer!

eregon:
Could you write here when the new release is done? Then I can close https://github.com/oracle/truffleruby/issues/1827.
(I subscribed to this repo's releases but I think it doesn't notify for tags)

SamSaffron:
Just cut a new release @eregon , is it working on truffle?

eregon:
Thanks!
Yes it works:
```
$ TRUFFLERUBYOPT="--jvm --polyglot" ruby -rmini_racer -ve 'ctx = MiniRacer::Context.new; p ctx.eval "1 + []"'
truffleruby 22.3.0-dev-4054469b, like ruby 3.0.3, GraalVM CE JVM [x86_64-linux]
"1"
```

The warnings when `--jvm --polyglot` are not passed aren't working correctly though, I'll look at that.

Issue #129, opened 2019-02-23
Accessing JS error object from MiniRacer::EvalError?


tisba:
I'm wondering how I can access the JavaScript error object from `MiniRacer::EvalError`.

I want to run script snippets from users and I want to report errors to them. We have some extended attributes on custom errors which I would need to access.

The only way I found to solve this is by wrapping the script before sending it to `MiniRacer::Context#eval` like this:

```ruby
wrapper = <<~JS
  try {
#{script}
  }
  catch(err) {
    var thrownError = err;
  }
JS

context.eval(wrapper)
```

With this I can access `thrownError`. But my issue is, that line numbers in backtraces are off. My current solution is to have a map that contains offsets accounting for the wrapper script.

Is there any other way to achieve this?

SamSaffron:
I think we need to fix our internals to account for the wrapper, can you give it a shot in a PR?

tisba:
I'm not sure I can follow.

I took a quick peek at https://github.com/discourse/mini_racer/blob/master/lib/mini_racer.rb#L170-L184 but could not spot anything that looks like a wrapper or exception handling.

In case you are referring to the the C code in https://github.com/discourse/mini_racer/blob/master/ext/mini_racer_extension/mini_racer_extension.cc, I have to pass (I don't understand it).

Just in case we have a misunderstanding: I'd like to access the thrown javascript error object:

```ruby
context.eval('throw new MyError("a message", { some: "meta data" });');
```

Is there a way to access the thrown error from miniracer?

SamSaffron:
I think you are talking about these lines: 

https://github.com/discourse/mini_racer/blob/723958f4da3cc6de01530b390cd4803dfbf7de7f/ext/mini_racer_extension/mini_racer_extension.cc#L736-L749

At the moment we send  a very rudimentary exception back so what we would want to do is look at the return result there and maybe convert the object to JSON and attach to Runtime error. It is somewhat tricky but we could provide more info for sure.

SamSaffron:
I think you are talking about these lines: 

https://github.com/discourse/mini_racer/blob/723958f4da3cc6de01530b390cd4803dfbf7de7f/ext/mini_racer_extension/mini_racer_extension.cc#L736-L749

At the moment we send  a very rudimentary exception back so what we would want to do is look at the return result there and maybe convert the object to JSON and attach to Runtime error. It is somewhat tricky but we could provide more info for sure.

tisba:
[...]

PR #264, opened 2022-09-21
pass exception text into javascript

seanmakesgames:
Our usage of mini_racer needs some way for exceptional cases in ruby to be able to be handled in javascript. Currently there is no way to identify what exception happened when it is caught by JS.

Also not sure what I do or do not need to do with handles and scopes here to make this string activity safe. Please let me know!

### DIFF
--- a/test/mini_racer_test.rb
+++ b/test/mini_racer_test.rb
@@ -877,7 +877,7 @@ raise FooError, "I like foos"
     error_message = "Actual Error Message"
     context.attach("a", proc{|a| raise error_message})
 
-    assert_equal error_message, context.eval("try { a(); \'no throw\' } catch (e) { e }")
+    assert_equal error_message, context.eval("try { a(); } catch (e) { e }")
   end
 
   def test_symbol_support
--- a/test/mini_racer_test.rb
+++ b/test/mini_racer_test.rb
@@ -872,6 +872,14 @@ raise FooError, "I like foos"
     end
   end
 
+  def test_capture_ruby_exception_message
+    context = MiniRacer::Context.new()
+    error_message = "Actual Error Message"
+    context.attach("a", proc{|a| raise error_message})
+
+    assert_equal error_message, context.eval("try { a(); \'no throw\' } catch (e) { e }")
+  end
+
   def test_symbol_support
     context = MiniRacer::Context.new()
     assert_equal :foo, context.eval("Symbol('foo')")
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -1281,7 +1281,8 @@ gvl_ruby_callback(void* data) {
 
     if(callback_data.failed) {
         rb_iv_set(parent, "@current_exception", result);
-        args->GetIsolate()->ThrowException(String::NewFromUtf8Literal(args->GetIsolate(), "Ruby exception"));
+        VALUE message = rb_funcall(result, rb_intern("message"), 0);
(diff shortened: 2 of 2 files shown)


seanmakesgames:
@bjfish Know why truffle test runs are failing here? (I don't know anything about truffleruby)

tisba:
For a second I was hoping this would also address https://github.com/rubyjs/mini_racer/issues/129 🙈

seanmakesgames:
@tisba If I read your issue correctly, then this PR might be an incremental start to what you are looking for. We’d also love to get a backtrace for the JS entry callstack as well as a typed Error object, but wanted to start with something less controversial and simpler

bjfish:
> Know why truffle test runs are failing here?
@eregon Could you review this?

seanmakesgames:
Any idea where to start with these?
Are there dev environment setup instructions somewhere?

Should I 'this feature doesn't work on truffle' the tests like some others?

Any feedback @SamSaffron ?

bjfish:
@eregon I suspect this would require calling `message`, string conversion if needed, and throwing in javascript (maybe with a helper method) similar to the implementation in this PR around: https://github.com/rubyjs/mini_racer/blob/master/lib/mini_racer/truffleruby.rb#L17-L23

test/mini_racer_test.rb
```diff
@@ -872,6 +872,14 @@ def test_pipe_leak
     end
   end
 
+  def test_capture_ruby_exception_message
+    context = MiniRacer::Context.new()
+    error_message = "Actual Error Message"
+    context.attach("a", proc{|a| raise error_message})
+
+    assert_equal error_message, context.eval("try { a(); } catch (e) { e }")
```
eregon, line 880: Isn't it rather weird that the return value of this is a JS exception, but the return value of this `context.eval` is just a string and not an exception? I think that's pretty confusing.
It seems like it should at least be wrapped into some MiniRacer exception or so.

eregon:
I think we need a review from maintainers first to decide the desired semantics.
For example what should be the return value of that test, what should a JS exception converted to Ruby return, etc.

Then I can push a commit to implement it for the GraalVM backend (fairly simple since it's all Ruby code).

seanmakesgames:
@eregon Yep. That implementation is a lot larger of a code change, though. @Fayti1703 is currently putting this together.
https://github.com/rubyjs/mini_racer/compare/master...Fayti1703:mini_racer:ex_msg_text

If those features are the minimum requirements for getting these in, we can make that happen. The changes in this PR are the minimum requirements for our usage to be able to 'discern in any way at all' which exception happened in ruby for error handling on the JS side.

Fayti1703:
The changes mentioned by seanmakesgames above are technically complete in the sense that they cause a JS-`Error` subclass to be thrown.

However one additional thing I'd like to do is pass through the Ruby error `VALUE` via the JS-Exception to rethrow it at the JS->Ruby boundary (instead of the current `@current_exception` storage, which only works for the first thrown error).

Currently still considering how best to do that, given the GC interactions.

I would be happy to create a PR with what I have so far, though.

eregon:
@SamSaffron Could you review this or ask another maintainer to review?

IMHO it's fine to get this in, although it seems of limited value if there is already a way to throw a proper JS Error.
I think a PR here to convert Ruby exceptions to JS exceptions would be welcome.

eregon:
I can do the corresponding changes on the mini_racer truffleruby backend, but would like this to be reviewed first.

SamSaffron:
Sean any chance you can test behavior of latest?

seanmakesgames:
> Sean any chance you can test behavior of latest?

Unfortunately not personally at the moment. Are you implying here though that this issue is fixed on latest? There might be a volunteer interested in checking in on it.

SamSaffron:
the whole implementation change radically, but I would hold off on testing till we are done stabilizing it.

PR #278, opened 2023-05-25
Bump required Ruby to 3.0, as 2.7 is EOL

tisba:
Ruby 2.7 has been EOL'd as of 2023-03-31 (see https://www.ruby-lang.org/en/downloads/branches/).

While removing references to 2.6/2.7 I noticed, that we did not test against Ruby 3.2 in GH Actions, so I added that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,10 +45,9 @@ jobs:
           - "macos-11"
           - "macos-12"
         ruby:
-          - "ruby-2.6"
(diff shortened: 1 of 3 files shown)


tisba:
hmm, not sure why the `test-truffleruby` should fail because of these changes. but since we're testing against head versions, I guess that's okay.

SamSaffron:
2.7 being EOL is such an end of an era... thanks for the PR @tisba

PR #279, opened 2023-05-25
Improved README & CHANGELOG

tisba:
I Fixed some typos and other nit-picky whitespace stuff in CHANGELOG.

I made a bit more changes to the README. I actually wanted to fix some typos and move the troubleshooting up based on issues opened by users having trouble installing mini_racer, but then I guess I got carried away and tried to make some more improvements 😁 

* fixed various typos, inconsistencies and extra whitespace
* I added a rubygem version "badge", because I saw quite a few bug reports not on the latest version, maybe this helps 🤞   Gem 
* I also moved the troubleshooting section up quite a bit, also in hopes that this helps to minimize reported bugs, that are easily fixed by using the troubleshooting guide 🤞🤞
* fixed a formatting error I added in https://github.com/rubyjs/mini_racer/pull/268 🤦 
* the memory soft limit example was broken. A non-existing `print()` was called, but it wasn't needed to showcase the feature anyway so I just dropped that
* in the examples, I switched to ECMAScript's [Template Literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) to show that mini_racer actually supports modern JS.
* I tried to add links to various things mentioned, like to the docs on how to load memory profiles into Chrome DevTools, etc.

Please let me know if that's too much in one go and I can split this up for easier review.

PR #280, opened 2023-05-26
Use latest bundler to get the fix of https://github.com/rubygems/rubygems/issues/6165

eregon:
Addresses https://github.com/rubyjs/mini_racer/pull/279#issuecomment-1563711199


eregon:
This works: https://github.com/eregon/mini_racer/actions/runs/5090437595/jobs/9149254323
The CI didn't run yet because I guess it needs an explicit approval, it would be nice to change that in settings so it just runs for contributors to the repo.

CvX (approved):

SamSaffron:
thanks @eregon !

PR #282, opened 2023-05-27
Adds dependabot config for updating Github actions

tisba:
We're not using a bunch of actions, but when they get updated, it would be nice to get a PR automatically once a week for us to review and merge.


lloeki:
Thanks @tisba, sounds like a healthy plan.

Issue #277, opened 2023-05-23
Node release catch-up plan


lloeki:
## Current status

- Jan 7, libv8-node `18.13.0.0` and `16.19.0.0` were built and pushed for all gem platforms for testing purposes
- Both are affected by the locale bug
- The locale bug is caused by an object file archive issue, and requires a package rebuild
- Fixes for the locale bug have landed on `node-16`, `node-17`, and `node-18` branches
- Fixes do not need a code change on `mini_racer`
- Node 17 and Node 18 require `mini_racer` code changes. PRs are open on [#271](https://github.com/rubyjs/mini_racer/pull/271) and [#261](https://github.com/rubyjs/mini_racer/pull/261)
- Node 18 OS requirements have increased a lot, notably dropping not-that-old macOS versions. I proposed having an intermediate `mini_racer` with `node-17` so as not to block these users.
- `aarch64-linux-musl` cross-compiling on CI is currently disabled due to some brokenness
- Node has had releases on 16 and 18 lines, notably Node 16 which had LTS security backports on 2023-03-28
- Node 19 has been released, with possibly challenging changes and updates to libv8, as is customary for non-LTS node.
- Node 20 (LTS) has been released, thus Node 19 won't have another bump

## Proposed plan

- [x] (1) build `node-16`, `node-17`, and `node-18` current branch tips locally, and publish the resulting gems to rubygems as `16.19.0.1`, `17.9.1.1`, and `18.13.0.1`.
  - [x] `16.19.0.1`
  - [x] `17.9.1.1`
  - [x] `18.13.0.1`

  This fixes those broken versions WRT the locale bug, and moves us closer to the "up-to-date goal" with builds we know are working.

- [x] (2) bump `mini_racer` dependency requirement to `~> 16.19.0.0` and release mini_racer `0.6.4`
  - [x] bump
  - [x] release

  This allows every user of current `mini_racer` to immediately move to a Node safer than 16.10.0 (and safest, short for one version)

- [x] (3) Merge [#271](https://github.com/rubyjs/mini_racer/pull/271) and release `mini_racer` `0.7.0`.
  - [x] merge
  - [x] release

  This allows every user of `mini_racer` to move to the latest node 17 even for OSes not supported by node 18.

- [x] (4) Merge [#261](https://github.com/rubyjs/mini_racer/pull/261) and release `mini_racer` `0.8.0`.
  - [x] merge
  - [x] release

[...]

PR #283, opened 2023-05-27
Update to libv8-node 19.x

lloeki:


Fayti1703:
Not sure if this is one issue or two issues, but that looks like a problem with managing V8 locks...
Going to try and debug this in a bit.

lloeki:
Awesome, thanks @Fayti1703

tisba:
Still WIP, but looks like we're changing the requirements for compiling mini_racer. If that's the case, we should probably also update the README (https://github.com/rubyjs/mini_racer#installation), which currently mentions:

> Note using v8.h and compiling MiniRacer requires a C++11 standard compiler, more specifically clang 3.5 (or later) or GCC 6.3 (or later).

I was also wondering if we should add a section to the README regarding supported macOS versions. I think @lloeki mentioned somewhere that the minimum supported macOS version changed with some v8 version.

Fayti1703:
I appear to be having a task-dependency issue when compiling locally:
```
Circular dependency detected:
TOP =>
compile =>
compile:x86_64-linux-gnu =>
compile:mini_racer_loader:x86_64-linux-gnu =>
copy:mini_racer_loader:x86_64-linux-gnu:3.0.2 =>
tmp/x86_64-linux-gnu/stage/lib/mini_racer_extension.so =>
lib/mini_racer_extension.so =>
copy:mini_racer_extension:x86_64-linux-gnu:3.0.2 =>
tmp/x86_64-linux-gnu/stage/lib/mini_racer_loader.so =>
copy:mini_racer_loader:x86_64-linux-gnu:3.0.2
```

This also occurs with `v0.8.0`, but not with `v0.6.4`.
(using `Bundler version 2.3.22` & `ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux-gnu]`)

Fayti1703:
Looks to be caused by 657a6ffcbfb84d2de34efe3204654a06b2d9dc3e...?

Fayti1703:
I've managed to pin down two issues here:
* `v8::platform::PumpMessageLoop` requires the Isolate to be locked.
   * crash in `test_webassembly`
   * **fixed by https://github.com/Fayti1703/mini_racer/commit/1e43a49ae550f8722071c92ca1248000ab49d6f3**
* Something deep in v8 is segfaulting when trying to read a snapshot -- specifically the constructor of `v8::internal::StringTableInsertionKey`. Unsure what the cause is at the moment.
  * crash in `test_snapshots_can_be_warmed_up_with_no_side_effects`
  * crash in `test_isolates_from_snapshot_dont_get_corrupted_if_the_snapshot_gets_warmed_up_or_GCed`

Fayti1703:
With a debug build of V8 (via a local libv8-node gem build), I get even *more* snapshot-related crashes due to invalid checksums. V8 says (for example):
```
#
# Fatal error in ../deps/v8/src/heap/read-only-spaces.cc, line 71
# Check failed: read_only_blob_checksum_ == snapshot_checksum (<unprintable> vs. 3847049538).
#
```

This smells like a V8 bug, since as far as I can tell MR doesn't touch the snapshot data whatsoever.

lloeki:
> This smells like a V8 bug

@Fayti1703 Did you by any chance attempt anything with node 20?

Fayti1703:
Not yet -- though judging from the CI runs on #284, there's similar behavior on that version as well.

I was planning on dumping out the checksum data on snapshot generation, trying to see if anything was off there (which would basically confirm it being a V8 issue), but I haven't found the time to do so yet.

lloeki:
A few weeks ago I made a half-hearted attempt to trace things using GDB reverse debugging time travel, but no dice.

Half-hearted because I was not very hopeful as GDB RD is quite limited + my RD skills are limited. A more potent RD is `rr`, but I'm not well versed with it, although I'm a bit more hopeful: when I got stuck with GDB RD falling short, someone managed to find stuff with `rr`.

Maybe it's time I ramp my RD+`rr` game up.

lloeki:
Yup, some vids are down below: https://rr-project.org

lib/mini_racer/version.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 module MiniRacer
-  VERSION = "0.8.0"
-  LIBV8_NODE_VERSION = "~> 18.16.0.0"
+  VERSION = "0.9.0"
+  LIBV8_NODE_VERSION = "~> 19.9.0.0"
```
mathieujobin, line 5: maybe we should skip non-LTS ?

tisba: You can find some more context to the plan we made a while ago here, @mathieujobin: https://github.com/rubyjs/mini_racer/issues/277

mathieujobin: Thanks, that indeed explains a lot.

Makes a lot of sense.

looking forward testing 0.10.0 with node20

lloeki:
I've moved to attempt things straight on node 20 here: https://github.com/rubyjs/mini_racer/pull/284

lloeki:

PR #284, opened 2023-05-27
Update to libv8-node 20.x

lloeki:
Depends on #283: uses `libv8-node-19` as a base, which should be merged first.

### DIFF
--- a/test/mini_racer_test.rb
+++ b/test/mini_racer_test.rb
@@ -1001,15 +1001,11 @@ raise FooError, "I like foos"
 
   def test_webassembly
     skip "TruffleRuby does not enable WebAssembly by default" if RUBY_ENGINE == "truffleruby"
-    log = File.open('testlog2', 'wb')
     context = MiniRacer::Context.new()
     context.eval("let instance = null;")
     filename = File.expand_path("../support/add.wasm", __FILE__)
     context.attach("loadwasm", proc {|f| File.read(filename).each_byte.to_a})
     context.attach("print", proc {|f| puts f})
-
-    log << 'a'
-    log.flush
 
     context.eval <<~JS
       WebAssembly
@@ -1023,22 +1019,12 @@ raise FooError, "I like foos"
         .then(i => { instance = i["instance"];})
         .catch(e => print(e.toString()));
     JS
-    log << 'b'
-    log.flush
 
     while !context.eval("instance") do
-      log << 'c'
-      log.flush
       context.isolate.pump_message_loop
-      log << 'd'
-      log.flush
     end
-    log << 'e'
-    log.flush
 
     assert_equal(3, context.eval("instance.exports.add(1,2)"))
-  ensure
-    log.close
   end
 
   class ReproError < StandardError
--- a/test/mini_racer_test.rb
+++ b/test/mini_racer_test.rb
@@ -999,56 +999,47 @@ raise FooError, "I like foos"
     assert_equal(v, 99)
   end
 
-# Fatal error in HandleScope::HandleScope
-# Entering the V8 API without proper locking in place
-# log has been seen as (in order):
-# abcdcdcdcdcdcdcdcdcdcdcdcdc
-# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
-# abcdcdcdcdcdcdcdcdcdcdcdcdc
-# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
-# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
-# cuts after c + cuts before d + never e => likely to be Isolate#pump_message_loop
-# def test_webassembly
-#   skip "TruffleRuby does not enable WebAssembly by default" if RUBY_ENGINE == "truffleruby"
-#   log = File.open('testlog2', 'wb')
-#   context = MiniRacer::Context.new()
-#   context.eval("let instance = null;")
-#   filename = File.expand_path("../support/add.wasm", __FILE__)
-#   context.attach("loadwasm", proc {|f| File.read(filename).each_byte.to_a})
-#   context.attach("print", proc {|f| puts f})
+  def test_webassembly
+    skip "TruffleRuby does not enable WebAssembly by default" if RUBY_ENGINE == "truffleruby"
+    log = File.open('testlog2', 'wb')
+    context = MiniRacer::Context.new()
+    context.eval("let instance = null;")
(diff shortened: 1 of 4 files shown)


test/mini_racer_test.rb
```diff
@@ -724,20 +734,23 @@ def test_can_dispose_context
     end
   end
 
-  def test_estimated_size
-    skip "TruffleRuby does not yet implement heap_stats" if RUBY_ENGINE == "truffleruby"
-    context = MiniRacer::Context.new(timeout: 5)
-    context.eval("let a='testing';")
+# Failure: expecting the isolate to have values for all the vals
+# {:total_physical_size=>835584, :total_heap_size_executable=>0, :total_heap_size=>1392640, :used_heap_size=>345736, :heap_size_limit=>1518338048}
```
lloeki, line 738: `total_heap_size_executable` seems to always be `0`

test/mini_racer_test.rb
```diff
@@ -724,20 +734,23 @@ def test_can_dispose_context
...
+#   stats = context.heap_stats
+#   # eg: {:total_physical_size=>1280640, :total_heap_size_executable=>4194304, :total_heap_size=>3100672, :used_heap_size=>1205376, :heap_size_limit=>1501560832}
+#   assert_equal(
+#     [:total_physical_size, :total_heap_size_executable, :total_heap_size, :used_heap_size, :heap_size_limit].sort,
+#     stats.keys.sort
+#   )
 
-    assert(stats.values.all?{|v| v > 0}, "expecting the isolate to have values for all the vals")
-  end
+#   File.open('testlog', 'wb') { |f| f << stats.inspect }
```
lloeki, line 751: Writing things to a file to debug

test/mini_racer_test.rb
```diff
@@ -986,33 +999,56 @@ def test_promise
...
+# log has been seen as (in order):
+# abcdcdcdcdcdcdcdcdcdcdcdcdc
+# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
+# abcdcdcdcdcdcdcdcdcdcdcdcdc
+# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
+# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
+# cuts after c + cuts before d + never e => likely to be Isolate#pump_message_loop
+# def test_webassembly
+#   skip "TruffleRuby does not enable WebAssembly by default" if RUBY_ENGINE == "truffleruby"
+#   log = File.open('testlog2', 'wb')
```
lloeki, line 1013: Writing things to a file to debug

test/mini_racer_test.rb
```diff
@@ -986,33 +999,56 @@ def test_promise
...
-  end
+# Fatal error in HandleScope::HandleScope
+# Entering the V8 API without proper locking in place
+# log has been seen as (in order):
+# abcdcdcdcdcdcdcdcdcdcdcdcdc
+# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
+# abcdcdcdcdcdcdcdcdcdcdcdcdc
+# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
+# abcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdcdc
+# cuts after c + cuts before d + never e => likely to be Isolate#pump_message_loop
```
lloeki, line 1010: This one seems to happen randomly after a few loops. Exact cause is yet unknown. Could be some timing, could be message loop doing no work but had some before, could be message loop having some work and none before, could be any other unknown thing really since I did not dig.

test/mini_racer_test.rb
```diff
@@ -539,38 +545,42 @@ def test_empty_isolate_is_valid_and_can_be_GCed
...
-  def test_isolates_from_snapshot_dont_get_corrupted_if_the_snapshot_gets_warmed_up_or_GCed
-    # basically tests that isolates get their own copy of the snapshot and don't
-    # get corrupted if the snapshot is subsequently warmed up
-    snapshot_source = <<-JS
-      function f() { return Math.sin(1); }
-      var a = 5;
-    JS
+# segfault
+# CFUNC  :init_with_snapshot
+# METHOD /Users/loic.nageleisen/Source/github.com/rubyjs/libv8-node/test/mini_racer/lib/mini_racer.rb:81 (Isolate#initialize)
```
lloeki, line 550: Initialising the isolate with a snapshot burns into flames. Note that it apparently doesn't reach `warmup!` so probably either unrelated to the above or indirectly related but for yet unknown, deeper reasons.

test/mini_racer_test.rb
```diff
@@ -448,29 +450,33 @@ def test_an_empty_snapshot_is_valid
...
 
-  def test_snapshots_can_be_warmed_up_with_no_side_effects
-    # shamelessly inspired by https://github.com/v8/v8/blob/5.3.254/test/cctest/test-serialize.cc#L792-L854
-    snapshot_source = <<-JS
-      function f() { return Math.sin(1); }
-      var a = 5;
-    JS
+# segfault
+# CFUNC  :warmup_unsafe!
+# METHOD /Users/loic.nageleisen/Source/github.com/rubyjs/libv8-node/test/mini_racer/lib/mini_racer.rb:452 (def warmup!)
```
lloeki, line 455: `warmup_unsafe!` indeed! Either we don't do something that we should to make it safe, or we pass something that was okay before and is not now.

```
    def warmup!(src)
      # we have to do something here
      # we are bloating memory a bit but it is more correct
      # than hitting an exception when attempty to compile invalid source
      begin
        ctx = MiniRacer::Context.new
        ctx.eval(@source)
        ctx.eval(src)
      rescue MiniRacer::RuntimeError => e
        raise MiniRacer::SnapshotError.new, e.message, e.backtrace
      end

      warmup_unsafe!(src)
    end
```

test/mini_racer_test.rb
lloeki: From: https://github.com/rubyjs/mini_racer/pull/283#issuecomment-1572735274 this was fixed by @Fayti1703 in https://github.com/Fayti1703/mini_racer/commit/1e43a49ae550f8722071c92ca1248000ab49d6f3, which I've cherry-picked.

lloeki:
Heh, just noticed `*-linux-gnu` has bundler electing to fetch the `ruby` platform gem instead of the binary ones, so spends time building libv8, which works for `x86_64` and probably also for `aarch64` if it did not get killed by GHA timeout.

lloeki:

PR #281, opened 2023-05-27
Remove unneeded files from release build

tisba:
I noticed that we include `.github/` and `bin/` when building. And other things :)

I think we can simplify this, by only including what we need, which is `lib/**/*` and `ext/**/*`. Additionally we probably want to continue to pack README, license, changelog and CoC.

**before** (released 0.6.4)

```
.
├── CHANGELOG
├── CODE_OF_CONDUCT.md
├── Dockerfile
├── Gemfile
├── LICENSE.txt
├── README.md
├── Rakefile
├── bin
│   ├── console
│   └── setup
├── ext
│   ├── mini_racer_extension
│   │   ├── Makefile
│   │   ├── extconf.rb
│   │   └── mini_racer_extension.cc
│   └── mini_racer_loader
│       ├── Makefile
│       ├── extconf.rb
│       └── mini_racer_loader.c
├── lib
│   ├── mini_racer
│   │   ├── truffleruby.rb
│   │   └── version.rb
│   ├── mini_racer.rb
│   ├── mini_racer_extension.bundle
│   └── mini_racer_loader.bundle
└── mini_racer.gemspec
```

**after** (build from `master`):

```
.
├── CHANGELOG
├── CODE_OF_CONDUCT.md
├── LICENSE.txt
├── README.md
├── ext
│   ├── mini_racer_extension
│   │   ├── Makefile
│   │   ├── extconf.rb
│   │   └── mini_racer_extension.cc
│   └── mini_racer_loader
│       ├── Makefile
│       ├── extconf.rb
│       └── mini_racer_loader.c
└── lib
    ├── mini_racer
    │   ├── truffleruby.rb
    │   └── version.rb
    ├── mini_racer.rb
    ├── mini_racer_extension.bundle
    └── mini_racer_loader.bundle
```

PR #285, opened 2023-06-02
gemspec: Exclude *.bundle (follow-up #281)

tisba:
Follow-Up for https://github.com/rubyjs/mini_racer/pull/281
Addresses https://github.com/rubyjs/mini_racer/pull/283#issuecomment-1572578821

I don't quite fully understand why this breaks, but building locally using `bundle exec rake compile` broke with #281. It works fine in CI though, which might be worth investigating. What worked for me to fix this (please have a look @Fayti1703) is to make sure `*.bundle` files are excluded. With this `bundle install && bundle exec rake` works (compiles and runs tests).

My experience with native extensions is limited. __I'd appreciate hints how to make local compilation consistent with CI__.

-cc @Fayti1703
-cc @lloeki

### DIFF
--- a/mini_racer.gemspec
+++ b/mini_racer.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
     "source_code_uri"   => "https://github.com/discourse/mini_racer/tree/v#{spec.version}",
   }
 
-  spec.files         = Dir["lib/**/*", "ext/**/*", "README.md", "LICENSE.txt", "CHANGELOG", "CODE_OF_CONDUCT.md"].reject { |fn| File.fnmatch?("**/*.bundle}", fn, File::FNM_EXTGLOB) }
+  spec.files         = Dir["lib/**/*.rb", "ext/**/*", "README.md", "LICENSE.txt", "CHANGELOG", "CODE_OF_CONDUCT.md"]
   spec.require_paths = ["lib"]
 
   spec.add_development_dependency "bundler"
--- a/mini_racer.gemspec
+++ b/mini_racer.gemspec
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
     "source_code_uri"   => "https://github.com/discourse/mini_racer/tree/v#{spec.version}",
   }
 
-  spec.files         = Dir["lib/**/*", "ext/**/*", "README.md", "LICENSE.txt", "CHANGELOG", "CODE_OF_CONDUCT.md"].reject { |fn| File.fnmatch?("**/*.{bundle,c,cc}", fn, File::FNM_EXTGLOB) }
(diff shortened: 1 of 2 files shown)


Fayti1703:
I've done some tests on my end... interestingly I see no change in behavior with the commits on this branch.
*However*, I have noticed that the problem doesn't even occur on a clean clone, which might explain why CI doesn't run into it.

mini_racer.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |spec|
     "source_code_uri"   => "https://github.com/discourse/mini_racer/tree/v#{spec.version}",
   }
 
-  spec.files         = Dir["lib/**/*", "ext/**/*", "README.md", "LICENSE.txt", "CHANGELOG", "CODE_OF_CONDUCT.md"]
+  spec.files         = Dir["lib/**/*", "ext/**/*", "README.md", "LICENSE.txt", "CHANGELOG", "CODE_OF_CONDUCT.md"].reject { |fn| File.fnmatch?("**/*.{bundle,c,cc}", fn, File::FNM_EXTGLOB) }
```
lloeki, line 24: Shouldn't this rather simply read `lib/**/*.rb`

lloeki: Also I think we need to keep the C files, otherwise the extension won't build once packaged as a gem.

tisba: erm, of course🤦 thanks

but one more possible item to improve CI builds, because it was all green.

tisba:
> I've done some tests on my end... interestingly I see no change in behavior with the commits on this branch.
> _However_, I have noticed that the problem doesn't even occur on a clean clone, which might explain why CI doesn't run into it.

I think that the clean & clobber tasks were not properly cleaning up everything. I added `{ext,lib}/**/*.{o,so,bundle}` and `pkg` to `CLEAN` so that with `rake clean clobber` you should get a clean environment again. I did some testing locally and it looks like a fresh clone. I think that should help somewhat.

tisba:
> Oh, this is interesting: Checking `Gemfile.lock`, my clean clone ends up with "PLATFORMS: x86_64-linux"[1](#user-content-fn-1-c8bfd2438f3f81ec1fc652fddc59ea85), where-as the existing worktree has a "PLATFORMS: ruby"[1](#user-content-fn-1-c8bfd2438f3f81ec1fc652fddc59ea85) declaration.
> 
> Can you verify if this is similar on your end?
> 
> ## Footnotes
> 1. In the original file, these are separated by newline + indentation. I've elected to replace this with a colon for readability. [↩](#user-content-fnref-1-c8bfd2438f3f81ec1fc652fddc59ea85) [↩2](#user-content-fnref-1-2-c8bfd2438f3f81ec1fc652fddc59ea85)

Not sure I follow 100%: For me it's always `PLATFORM: arm64-darwin-22` in `Gemfile.lock` when it's generated.

SamSaffron:
this is great, thanks.

PR #287, opened 2023-06-05
CI: Fix "::set-output" deprecation

tisba:
Looks like Github is postponing the removal, but IMO we should still address this, see https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/:

> Our telemetry shows significant usage of these commands. Given the number of impacted customers we are postponing the removal.

SCR-20230605-iccc


tisba:
hmm, we might want to set a reasonable timeout on our jobs maybe. Looks like the macOS truffleruby job hang for 6 hours without making progress 🤷 

In any case, has (most likely) nothing to do with this changeset.

lloeki (approved):

tisba:
Feel free to merge @lloeki (I can't).

PR #299, opened 2024-04-08
Update to libv8-node 21.x

lloeki:
Tentative PR to examine results with latest libv8


lloeki:
This does not crash like node 19 and 20, libv8 might have fixed something upstream.

lloeki:

PR #306, opened 2024-07-23
Update to libv8-node 22.5.1.0

lloeki:


lloeki:
As we saw with Sam, the failing test appears to be from V8 optimising an empty function out. I'll fix that by making it do something (same way I did for another similar test).

Also I fat fingered `LIBV8_VERSION` in `version.rb` :(

And since I want to fix this one too https://github.com/rubyjs/libv8-node/pull/36 I'll certainly do a .1 release (which will be automatically picked up given the on-purpose libv8-node version requirement in mini_racer) once they're done.

tisba:
I'm not sure why the `test_timeout` uses a `Promise` to begin with, but the timeout is triggered (correctly IMO), by just `while(true) {}`. At least in my testing, this makes the test pass (tested on aarch64 and Darwin 23.5.0 arm64):

```diff
diff --git i/test/mini_racer_test.rb w/test/mini_racer_test.rb
index 6b139fd..f1fd1b9 100644
--- i/test/mini_racer_test.rb
+++ w/test/mini_racer_test.rb
@@ -1044,12 +1044,8 @@ def test_timeout
     context = MiniRacer::Context.new(timeout: 500, max_memory: 20_000_000)
     assert_raises(MiniRacer::ScriptTerminatedError) do
       context.eval <<~JS
-        var doit = async() => {
-        while (true)
-          await new Promise(resolve => resolve())
-        }
-        doit();
-        JS
+        while(true) {}
+      JS
     end
   end
```

SamSaffron:
Cool made https://github.com/rubyjs/mini_racer/pull/308

dropping Ruby 3.0 as well while at it.

PR #310, opened 2024-08-15
Fix "can't alloc thread" exception on exit

bnoordhuis:
It's no longer possible to create new threads once Ruby's main thread has exited.

<hr>

Specifically, it works around this: https://github.com/ruby/ruby/blob/abc04e898b/thread.c#L900-L902


SamSaffron:
Thanks Ben.

PR #311, opened 2024-08-20
Fix segfault, disable V8's thread-isolated allocator

bnoordhuis:
Fixes: https://github.com/rubyjs/mini_racer/issues/300

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.cc
+++ b/ext/mini_racer_extension/mini_racer_extension.cc
@@ -3,9 +3,14 @@
 #include <ruby/thread.h>
 #include <ruby/io.h>
 #include <ruby/version.h>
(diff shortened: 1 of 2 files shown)


bnoordhuis:
This is against V8 12.x. Is there a specific branch I should be targeting instead of main? libv8-node-22?

SamSaffron:
Sorry for the short reply, getting late :)

You will need to also edit version.rb to target the right version, we
reverted that when we were debugging the segfaults

Is this patch going to be easy to maintain?

bnoordhuis:
I'll take a closer look tomorrow why `rake test` segfaults (it's not a pkey thing) but if it's not something straightforward, I'll just PR libv8-node instead and disable PKRU in V8.

bnoordhuis:
I'll PR libv8-node instead, closing.

PR #315, opened 2024-09-17
Disable strict aliasing

bnoordhuis:
Harmonize C++ compiler flags with Node.js (-fno-rtti, -fno-exceptions) and, critically, disable strict aliasing, because V8 is not the least bit strict-aliasing safe.

Upstream turned it off in https://github.com/nodejs/node/pull/54339, like it always should have been, and I suspect the mismatch may be the cause of segfaults that show up in production.

~~Remove -fpermissive, it's not needed and something to discourage.~~

### DIFF
--- a/ext/mini_racer_extension/extconf.rb
+++ b/ext/mini_racer_extension/extconf.rb
@@ -19,6 +19,9 @@ $CXXFLAGS += " -rdynamic" unless $CXXFLAGS.split.include? "-rdynamic"
 $CXXFLAGS += " -fPIC" unless $CXXFLAGS.split.include? "-rdynamic" or IS_DARWIN
 $CXXFLAGS += " -std=c++17"
 $CXXFLAGS += " -fpermissive"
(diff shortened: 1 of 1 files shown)


bnoordhuis:
macos failures:

> curl: (28) Failed to connect to nodejs.org port 443 after 150033 ms: Couldn't connect to server

i.e., network outage. It's trying to build libv8-node from source, is that expected?

musl failures:

>  /usr/lib/gcc/aarch64-alpine-linux-musl/13.2.1/../../../../aarch64-alpine-linux-musl/bin/ld: cannot find /usr/local/bundle/gems/libv8-node-22.7.0.4-aarch64-linux/vendor/v8/aarch64-linux-musl/libv8/obj/libv8_monolith.a: No such file or directory

Pre-existing, same build error on main. Started with commit 9ef1982, I think.

tisba:
musl error is most likely because of https://github.com/rubyjs/libv8-node/issues/60

SamSaffron (approved):

PR #328, opened 2025-01-08
Fix TruffleRuby backend after #325

eregon:
See https://github.com/rubyjs/mini_racer/pull/325#issuecomment-2577491290 and after for context.
All tests pass now, so this should fix the CI.

### DIFF
--- a/lib/mini_racer.rb
+++ b/lib/mini_racer.rb
@@ -75,14 +75,12 @@ module MiniRacer
 
       if String === file_or_io
         f = File.open(file_or_io, "w")
-	implicit = true
+        implicit = true
       else
         f = file_or_io
       end
 
-      if !(File === f)
-        raise ArgumentError, "file_or_io"
-      end
+      raise ArgumentError, "file_or_io" unless File === f
 
       f.write(heap_snapshot())
     ensure
--- a/lib/mini_racer/shared.rb
+++ b/lib/mini_racer/shared.rb
@@ -353,7 +353,7 @@ module MiniRacer
         ctx = MiniRacer::Context.new
         ctx.eval(str)
       rescue MiniRacer::RuntimeError => e
-        raise MiniRacer::SnapshotError.new, e.message, e.backtrace
+        raise MiniRacer::SnapshotError, e.message, e.backtrace
       end
 
       @source = str
@@ -371,7 +371,7 @@ module MiniRacer
         ctx.eval(@source)
         ctx.eval(src)
       rescue MiniRacer::RuntimeError => e
-        raise MiniRacer::SnapshotError.new, e.message, e.backtrace
+        raise MiniRacer::SnapshotError, e.message, e.backtrace
       end
 
       warmup_unsafe!(src)
--- a/lib/mini_racer/shared.rb
+++ b/lib/mini_racer/shared.rb
@@ -1,0 +1,380 @@
+# This code used to be shared in lib/mini_racer.rb
+# but was moved to the extension with https://github.com/rubyjs/mini_racer/pull/325.
+# So now this is effectively duplicate logic with C/C++ code.
+# Maybe one day it can be actually shared again between both backends.
+
+module MiniRacer
+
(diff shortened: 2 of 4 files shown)


lib/mini_racer.rb
```diff
@@ -75,14 +75,12 @@ def write_heap_snapshot(file_or_io)
 
       if String === file_or_io
         f = File.open(file_or_io, "w")
-	implicit = true
+	      implicit = true
```
bnoordhuis, line 78: Accidental whitespace change?

lib/mini_racer/shared.rb
```diff
@@ -0,0 +1,380 @@
+# This code used to be shared in lib/mini_racer.rb
+# but was moved to the extension with https://github.com/rubyjs/mini_racer/pull/325.
+# So now this is effectively duplicate logic with C/C++ code.
+# Maybe one day it can be actually shared again between both backends.
+
+module MiniRacer
+
+  MARSHAL_STACKDEPTH_DEFAULT = 2**9-2
+  MARSHAL_STACKDEPTH_MAX_VALUE = 2**10-2
```
bnoordhuis, line 9: FWIW, these are gone; the C code accepts `marshal_stack_depth` but otherwise ignores it.

lib/mini_racer/shared.rb
```diff
@@ -0,0 +1,380 @@
...
+
+      @disposed = false
+
+      @callback_mutex = Mutex.new
+      @callback_running = false
+      @thread_raise_called = false
+      @eval_thread = nil
+
+      # defined in the C class
+      init_unsafe(isolate, snapshot)
```
bnoordhuis, line 134: init_unsafe no longer exists

lib/mini_racer/shared.rb
```diff
@@ -0,0 +1,380 @@
...
+    end
+
+    def isolate
+      return @isolate if @isolate != false
+      # defined in the C class
+      @isolate = create_isolate_value
+    end
```
bnoordhuis, line 141: Isolates no longer exist (or rather, are not exposed in the Ruby API)

bnoordhuis (commented):
Left some comments.

lib/mini_racer.rb
eregon: Ah there must be a tab in there, it looks misaligned in my editor, will fix to use spaces.

eregon: Fixed.

lib/mini_racer/shared.rb
eregon: It does, it's defined in `lib/mini_racer/truffleruby.rb`.

lib/mini_racer/shared.rb
eregon: Right, I took a quick look but it's not so easy to remove so that could be done separately and I think best to get the CI green first (e.g. to avoid regressions).

bnoordhuis: README update in #331. The changelog was Sam's work, I think?

test/mini_racer_test.rb
```diff
@@ -1040,6 +1046,9 @@ def test_forking
   end
 
   def test_poison
+    if RUBY_ENGINE == "truffleruby"
+      skip "TruffleRuby uses some extra JS code when creating/using a Contex which seems to trigger the poison"
```
bnoordhuis, line 1050:
```suggestion
      skip "TruffleRuby uses some extra JS code when creating/using a Context which seems to trigger the poison"
```

bnoordhuis (approved):
I'm probably not the most qualified person to review but LGTM at a quick glance. Thanks for doing this, @eregon.

SamSaffron (approved):

SamSaffron:
thank you so much @eregon for the push here, I know how much extra work out of nowhere there was but our upgrade of our v8 integration was critical for the long term survival of MiniRacer, so happy to bring back full support

PR #329, opened 2025-01-09
Add failing test for function cloning

SamSaffron:


SamSaffron:
@bnoordhuis I just tried latest on Discourse and it is failing due to this test:

```
  def test_function_cloning
    message = nil
    context.attach("rails.logger.warn", proc { |msg| message = msg })
    context.eval <<~JS
               console = {
                 prefix: "[PrettyText] ",
                 log: function(...args){ rails.logger.info(console.prefix + args.join(" ")); },
               }
             JS

    context.eval("console.log('Hello', 'World!')")
    assert_equal "[PrettyText] Hello World!", message
  end
```

bnoordhuis:
Typo? It says `rails.logger.warn` on one line and `rails.logger.info` on another.

SamSaffron:
@bnoordhuis sorry about the typo fixed it, but test is still broken due to clone issue

test/mini_racer_test.rb
```diff
@@ -1067,9 +1065,26 @@ def test_regexp_string_iterator
...
   end
+
+  def test_function_cloning
+    message = nil
+    context = MiniRacer::Context.new
+    context.attach("log", proc { |msg| message = msg })
+    context.eval <<~JS
+       console = {
+         log: function(...args){ log(args.join(" ")); },
+       }
```
bnoordhuis, line 1084: I see what the problem is: the result of this statement is the `log` function object. The test passes when changed to this:
```suggestion
       console = {
         log: function(...args){ log(args.join(" ")); },
       }
       undefined
```
Let me think what a good solution is.

bnoordhuis: For completeness sake, another workaround is to make the property non-enumerable:
```suggestion
       console = {}
       Object.defineProperty(console, "log", {value: function(...args){ log(args.join(" ")); }})
```

bnoordhuis: For posterity, I think there are 4 options:

1. do nothing and let the error bubble up (current situation)
2. detect the error and return something else, like an empty object or a marker
3. deep-copy the object and replace or drop function properties
4. ditto but only shallow-copy the object, i.e., fix up only one level

3 is O(n) time and space, 4 is only a partial fix, so I guess 2 is the best option.

PR #330, opened 2025-01-09
Handle DataCloneError serialization errors

bnoordhuis:
Instead of trying to detect upfront if an object can be serialized, just do it and see if it throws a DataCloneError.

Unexpected DataCloneErrors were of course already being handled but the new approach does it in a centralized and generic manner.

Failed serialization now passes an object with a single "error" string property back to Ruby land.

Fixes the failing test from https://github.com/rubyjs/mini_racer/pull/329


SamSaffron:
I love that it is simpler, it remains a breaking change vs old behavior which would handle partial clones but I think this is probably ok, when we release this we can do a major version bump, will try this on Discourse today

PR #335, opened 2025-01-11
Convert deserialized strings to UTF-8

bnoordhuis:


test/mini_racer_test.rb
```diff
@@ -1103,5 +1103,11 @@ def test_string_encoding
     assert_equal "ok", context.eval("'ok'".encode("ISO-8859-1"))
     assert_equal "ok", context.eval("'ok'".encode("ISO8859-1"))
     assert_equal "ok", context.eval("'ok'".encode("UTF-16LE"))
+    assert_equal Encoding::UTF_8, context.eval("'ok'").encoding
+    assert_equal Encoding::UTF_8, context.eval("'ok\\uD800\\uDC00'").encoding
+    if RUBY_ENGINE != "truffleruby"
+      # unmatched surrogate pair, cannot be converted by ruby
+      assert_equal Encoding::UTF_16LE, context.eval("'ok\\uD800'").encoding
```
bnoordhuis, line 1110: @SamSaffron the alternative is to let the exception that ruby raises bubble up

SamSaffron: Yeah not sure if there is an easy way of winning here, lets try this out first.

SamSaffron (approved):

PR #336, opened 2025-01-18
Add property filter fallback mode to serializer

bnoordhuis:
For a JS value that cannot be serialized as-is by V8's serializer, clone and filter its properties, then try again. With this change, serialization should be possible for practically all JS values.

### DIFF
--- a/test/mini_racer_test.rb
+++ b/test/mini_racer_test.rb
@@ -1089,11 +1089,19 @@ class MiniRacerTest < Minitest::Test
 
   def test_function_property
     context = MiniRacer::Context.new
-    expected = {
-      "m" => {"1" => 2, "3" => 4},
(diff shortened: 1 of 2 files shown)


test/mini_racer_test.rb
```diff
@@ -1083,19 +1083,34 @@ def test_regexp_string_iterator
...
     assert_equal expected, context.eval("'abc'.matchAll(/./g)")
   end
 
   def test_function_property
     context = MiniRacer::Context.new
     if RUBY_ENGINE == "truffleruby"
-      expected = {"x" => 42}
+      expected = {
+        "m" => {1 => 2, 3 => 4},
+        "s" => {},
```
bnoordhuis, line 1095: @eregon fyi - seems like a bug

SamSaffron (approved):

PR #339, opened 2025-01-21
Add Ruby 3.4 to build

tisba:


tisba:
I'll also send a PR using Github's arm64 runner (public preview): https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/. This should speed up our build times quite a bit and simplifies the workflow.

PR #341, opened 2025-01-22
Use GH hosted ARM64 linux runners

tisba:
GitHub hosted Linux ARM64 runners are now available, see https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/.

Let's see if we can use them, simplify our build workflow and make it faster 🚀 

Okay, that's A LOT faster. Almost all jobs done in under one minute. Especially the ones that used to run under emulation, where VERY slow, like over 15min.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,22 +90,17 @@ jobs:
           - "3.2"
           - "3.3"
           - "3.4"
-        platform:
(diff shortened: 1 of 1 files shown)


tisba:
looks like the public preview of the ARM runners is helplessly overloaded 😢

PR #342, opened 2025-02-04
Update to V8 12.9

bnoordhuis:
Removes context.idle_notification because the V8 API it uses has been removed with no real replacement.

V8 suggests to use Isolate::MemoryPressureNotification() instead but that's not really the same thing and overlaps significantly with context.low_memory_notification.

Remove the stale object marshalling section from the README.

<hr>

Needs a libv8-node release from its node-23 branch first

### DIFF
--- a/ext/mini_racer_extension/extconf.rb
+++ b/ext/mini_racer_extension/extconf.rb
@@ -19,7 +19,7 @@ $CXXFLAGS += " -Wall" unless $CXXFLAGS.split.include? "-Wall"
 $CXXFLAGS += " -g" unless $CXXFLAGS.split.include? "-g"
 $CXXFLAGS += " -rdynamic" unless $CXXFLAGS.split.include? "-rdynamic"
 $CXXFLAGS += " -fPIC" unless $CXXFLAGS.split.include? "-rdynamic" or IS_DARWIN
-$CXXFLAGS += " -std=c++17"
+$CXXFLAGS += " -std=c++20"
 $CXXFLAGS += " -fpermissive"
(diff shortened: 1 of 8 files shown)


ext/mini_racer_extension/extconf.rb
```diff
@@ -19,7 +19,7 @@
 $CXXFLAGS += " -g" unless $CXXFLAGS.split.include? "-g"
 $CXXFLAGS += " -rdynamic" unless $CXXFLAGS.split.include? "-rdynamic"
 $CXXFLAGS += " -fPIC" unless $CXXFLAGS.split.include? "-rdynamic" or IS_DARWIN
-$CXXFLAGS += " -std=c++17"
+$CXXFLAGS += " -std=c++20"
```
tisba, line 22: Do we need to update the README for compiler requirements too? https://github.com/rubyjs/mini_racer?tab=readme-ov-file#installation

bnoordhuis: Good point, those are way out of date. Updated with the compiler recommendations from node's BUILDING.md

SamSaffron:
any ideas why all the tests are failing? I think it is struggling with bundle for some reason.

bnoordhuis:
@SamSaffron 

> Could not find gem 'libv8-node (~> 23.6.1.0)', which is required by gem

You can do libv8-node releases now, right? If you make one, I'll restart CI and things should - fingers crossed - be green.

SamSaffron:
Will do tomorrow

PR #345, opened 2025-03-13
Handle termination exceptions during serialization

bnoordhuis:
Don't abort when the ruby thread calls `context.stop` right when the V8 thread is serializing its response to a request. Temporarily cancel the termination exception, retry serialization, and reraise the exception.


bnoordhuis:
The musl failures are partially the old "libv8-node not found" build errors but also partially this one that I hadn't seen before:
```
TZInfo::DataSourceNotFound: No source of timezone data could be found
```
Not caused by this PR, at least.

SamSaffron (approved):

Issue #348, opened 2025-03-31
Large integers are incorrectly serialized as floats


SamSaffron:
```
  def test_large_integer
    ctx = MiniRacer::Context.new
    big_int = 10_000_000_001
    ctx.attach("test", proc { big_int })

    result = ctx.eval("test()")

    assert_equal(result.class, big_int.class)
    assert_equal(result, big_int)
  end

```

The following test fails, even though serde has bigint support so somehow we are not properly converting stuff to bigints where we should be.

bnoordhuis:
10_000_000_001 is bigger than INT32_MAX so it's converted to a double, i.e., a JS number.

It's possible to send it over as a BigInt instead but I don't know if you actually want to do that because it's a breaking change on the JS side: `x + y` where `x` is a number starts failing with an exception when `y` changes from number to bigint.

Maybe another option is to deserialize doubles that are round numbers to fixnums/bignums instead of floats. I don't know if and how much of a breaking change that is for mini_racer. It's definitely observable.

SamSaffron:
oh yeah this is very tricky indeed, cause it is boundary issue, if the number is way bigger we have no choice but to bigint it, but on the boundary JS quirks end up hurting us: 

![Image](https://github.com/user-attachments/assets/71001de6-281c-45ae-a471-85394640f54e)

I guess the rabbit hole leads us to this failing test: 

```
  def test_large_integer
    ctx = MiniRacer::Context.new
    big_int = 123_456_789_123_456_789
    ctx.attach("test", proc { big_int })

    result = ctx.eval("test()")

    assert_equal(result.class, big_int.class)
    assert_equal(result, big_int)
  end
```

bnoordhuis:
> I guess the rabbit hole leads us to this failing test

Yes, Ruby Integers outside the range `-2**53` to `2**53` are currently not serializable because JS numbers are doubles.

I can make it so those are converted to bigints on the JS side instead of throwing the current, fairly opaque DataCloneError. Good idea?

SamSaffron:
yeah, I think it is better, cause otherwise it is very confusing, its quite rare, no rush here for sure.

bnoordhuis:
#350

PR #349, opened 2025-03-31
Convert round doubles to fixnums instead of floats

bnoordhuis:
Fixes: https://github.com/rubyjs/mini_racer/issues/348


SamSaffron:
thanks Ben!

PR #350, opened 2025-04-07
Convert big ruby integers to js bigints

bnoordhuis:
Numbers outside the range that double-precision floating-point numbers can represent exactly are now converted to JS BigInts instead of being rejected outright. Said range corresponds to `Number.MIN_SAFE_INTEGER` and `Number.MAX_SAFE_INTEGER` on the JS side.

Refs: https://github.com/rubyjs/mini_racer/issues/348


SamSaffron:
Thanks Ben!

Issue #359, opened 2025-07-16
Error installing mini_racer v0.19.0 on macOS 15.7.2 x86


afdev82:
I am trying to install mini_racer v0.19.0 on a MacBook Pro and I am getting the following error:

```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /Users/antonio.facciolo/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/mini_racer-0.19.0/ext/mini_racer_extension
/Users/antonio.facciolo/.rbenv/versions/3.4.4/bin/ruby extconf.rb
checking for -lpthread... yes
checking for -lobjc... yes
checking for whether -Wl,--exclude-libs=ALL  is accepted as LDFLAGS... no
creating Makefile

current directory: /Users/antonio.facciolo/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/mini_racer-0.19.0/ext/mini_racer_extension
make DESTDIR\= sitearchdir\=./.gem.20250716-48170-u24r96 sitelibdir\=./.gem.20250716-48170-u24r96 clean

current directory: /Users/antonio.facciolo/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/mini_racer-0.19.0/ext/mini_racer_extension
make DESTDIR\= sitearchdir\=./.gem.20250716-48170-u24r96 sitelibdir\=./.gem.20250716-48170-u24r96
compiling mini_racer_extension.c
In file included from mini_racer_extension.c:13:
./serde.c:101:27: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
    n = next_power_of_two(n);
        ~~~~~~~~~~~~~~~~~ ^
./serde.c:111:14: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
    b->cap = n;
           ~ ^
./serde.c:398:29: warning: implicit conversion loses integer precision: 'uint64_t' (aka 'unsigned long long') to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
        des_object_ref(arg, u);
        ~~~~~~~~~~~~~~      ^
[...]

PR #361, opened 2025-07-24
[#359] extension: fix typing

gi:
This updates the typing to be more consistent and avoid implicit conversion warnings/errors when compiling on darwin/macOS.


gi:
The `linux-ubuntu-24.04-arm - ruby-3.x - musl` jobs are failing on `main`, so these errors do not seem to have anything to do with these changes.

tisba:
musl builds are known to be broken currently, don't worry too much about them.

gi:
> musl builds are known to be broken currently, don't worry too much about them.

@tisba Are you good merging this then? 

The changes fix the compilation issues I see on my machine. It aligns the types of variables declared in function bodies and in called function signatures or with the Ruby C API.

ext/mini_racer_extension/mini_racer_extension.c
```diff
@@ -651,7 +651,7 @@ static int serialize(Ser *s, VALUE v)
     return serialize1(s, rb_hash_new(), v);
 }
 
-static struct timespec deadline_ms(int ms)
+static struct timespec deadline_ms(int64_t ms)
```
gi, line 654: Avoids issues when `int` is 32-bits.

ext/mini_racer_extension/serde.c
```diff
@@ -54,13 +55,18 @@ typedef struct Ser {
 static const uint8_t the_nan[8] = {0,0,0,0,0,0,0xF8,0x7F}; // canonical nan
 
 // note: returns |v| if v in [0,1,2]
-static inline uint32_t next_power_of_two(uint32_t v) {
+static inline size_t next_power_of_two(size_t v) {
```
gi, line 58: widen to handle 64-bit `size_t` arguments to function call.

ext/mini_racer_extension/serde.c
```diff
@@ -240,24 +246,26 @@ static void ser_num(Ser *s, double v)
 }
 
 // ser_bigint: |n| is in bytes, not quadwords
-static void ser_bigint(Ser *s, const uint64_t *p, size_t n, int sign)
+static void ser_bigint(Ser *s, const void *p, size_t stride, size_t n, int sign)
```
gi, line 249: Updated to handle both 32- and 64-bit input pointers given two invocations in the code.

tisba:
> @tisba Are you good merging this then?

I'm not really capable of judging if your changes are okay not, sorry 🫣 Builds are green (except the arm+musl ones which are okay to fail currently). I cannot help to confirm if your initial problem got fixed (build issue on macOS, see #359), because I cannot reproduce it.

Maybe @bnoordhuis or @SamSaffron could have a look?

ext/mini_racer_extension/serde.c
```diff
@@ -324,26 +332,26 @@ static void ser_object_begin(Ser *s)
...
-static void ser_array_begin(Ser *s, uint32_t count)
+static void ser_array_begin(Ser *s, size_t count)
 {
     w_byte(s, 'A');     // 'A'=dense, 'a'=sparse
     w_varint(s, count); // element count
 }
 
 // |count| is the element count
-static void ser_array_end(Ser *s, uint32_t count)
+static void ser_array_end(Ser *s, size_t count)
```
gi, line 354: This is the same argument as ser_object_end: https://github.com/rubyjs/mini_racer/pull/361/files#r2234665993.

> The one invocation of this function [(that is not a constant 2)] in serialize1 passes in a size_t value. It's passed to w_varint which is declared as void w_varint(Ser *s, uint64_t v) with a uint64_t.
> 
> Thus, it makes sense to align this function's parameter type with the invocation argument's type of size_t which can easily be widened to a 64-bit uint64_t if size_t happens to be 32 bits.

ext/mini_racer_extension/serde.c
```diff
@@ -569,7 +577,7 @@ static int des1(char (*err)[64], const uint8_t **p, const uint8_t *pe,
             return bail(err, "negative zero bigint");
         if (pe-*p < (int64_t)u)
             goto too_short;
-        des_bigint(arg, *p, u, 1-2*t);
+        des_bigint(arg, *p, u, (int) (1-2*t));
```
gi, line 580: The parameter only designates the sign here: `1` or `-1`, but the type of `t` is `uint64_t`. Thus, instead of making the parameter of `des_bigint` unnecessarily wide, it's easy enough to just cast it down explicitly to avoid any potential compiler warnings that can turn into errors.

gi:
@bnoordhuis I commented on the changes in the files diff. It might be easier to address each there; however, I will briefly respond here.

> * an object reference is not `long` because it can't be < 0 or > UINT32_MAX (and if `long` is 32 bits it's doubly wrong because values _can_ be > INT32_MAX)

Ruby's C API `rb_ary_entry` uses the type `long` for the offset, so it makes sense to forward this type requirement to the callers in this library.

See https://github.com/rubyjs/mini_racer/pull/361/files#r2233122575.

> * changing next_power_of_two's argument type from `uint32_t` to `size_t` and all the changes that flow from that makes no sense because arrays in the wire format cannot be > UINT32_MAX; it merely complicates the function and adds dead code

The sole invocation of this function passes in an argument of type size_t, so this function should declare the parameter of the same type to simplify things.

See https://github.com/rubyjs/mini_racer/pull/361/files#r2233124981.

> * the ser_bigint changes look unnecessary; I'm not sure I understand what it's trying to fix

See https://github.com/rubyjs/mini_racer/pull/361/files#r2233614647.

bnoordhuis:
The ser/de function types mirror the sizes of fields in the wire format. Replacing e.g. `uint32_t` with `size_t` opens up the possibility of passing values > UINT32_MAX and then... well, I don't know exactly what would happen without looking at the code but it most assuredly would go wrong.

If you want to squelch those 64-to-32 warnings, you can narrow-cast at the call sites. `r_varint` and `w_varint` currently operate on uint64s but in practice<sup>1</sup> wire most format values (with exceptions for e.g. bigints and typed arrays) are 32 bits unsigned values, so you could probably avoid most warnings by introducing `r_varint32` and `w_varint32`.

Having said that, unless you have a good working knowledge of the wire format, there's a pretty good chance you'll break more than you fix, so tread carefully.

<sup>1</sup> the spec, what there is of it, doesn't say so but in practice V8's serializer only writes 32 bits values in most places

PR #363, opened 2025-08-02
Fix object identity bug in Ruby-to-JavaScript conversion

benjaminwood:
> [!WARNING]
> C is not my area of expertise. Vibe code content warning 😅 - please review as such! Thank you!

This fixes a bug where separate Ruby objects with identical content were incorrectly sharing the same JavaScript object reference during serialization.

Problem:
- Ruby hashes like `[{"value" => 42}, {"value" => 42}, {"value" => 42}]` were becoming the same JavaScript object reference
- Mutating one object affected all identical objects
- This happened because serialization used Ruby's content-based hash equality instead of object identity for reference tracking

Solution:
- Modified `serialize1()` in `mini_racer_extension.c` to use `rb_obj_id(v)` as the hash key instead of the object itself (v)
- This ensures only truly identical Ruby objects (same memory address) share JavaScript references, preserving circular reference handling
- Separate Ruby objects now correctly become separate JavaScript objects

Changes:
- ext/mini_racer_extension/mini_racer_extension.c: Use object identity for refs
- test/test_object_identity_fix.rb: Comprehensive regression tests

Tests:
- New regression tests verify both separate and shared object scenarios
- Validates fix for nested objects and circular references

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.c
+++ b/ext/mini_racer_extension/mini_racer_extension.c
@@ -527,52 +527,58 @@ static int serialize1(Ser *s, VALUE refs, VALUE v)
         return -1;
     switch (TYPE(v)) {
     case T_ARRAY:
-        id = rb_hash_lookup(refs, v);
-        if (NIL_P(id)) {
-            n = RARRAY_LENINT(v);
-            i = rb_hash_size_num(refs);
-            rb_hash_aset(refs, v, LONG2FIX(i));
-            ser_array_begin(s, n);
-            for (i = 0; i < n; i++)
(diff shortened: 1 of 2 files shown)


SamSaffron:
Thanks so much, will leave @bnoordhuis to have a review, love the tests.

benjaminwood:
> love the tests

When vibe coding OSS, the least you can do is write some tests 😄 

(Hey, that rhymed!)

bnoordhuis:
Expectation management: I'm on vacation and won't be back for another ~1.5 week.

bnoordhuis (approved):
LGTM, thanks.

Issue #357, opened 2025-07-07
catch ruby exception (raised in attached function) within JS: change in behaviour


pascalbetz:
Hi Mini Racer team.

We've run into an issue with MiniRacer and are wondering if we were relying on a undocumented feature or if this is a bug in more recent versions.

We used to be able to catch exception thrown in attached ruby functions inside a JS `catch` block. The ruby exception was "converted" to a JS exception. 

Running this ruby script

```ruby
require "mini_racer"
context = MiniRacer::Context.new
context.attach("externalFunction", proc { raise "raised in ruby"} )
result = context.eval(<<~JS)
  try {
    externalFunction()
  } catch(error) {
   "catch in JS"
  }
JS
```

gives me different results starting with 0.17.0.pre6

**mini_racer-0.17.0.pre5** (and in all version since at least 0.12.0)
```
"catch in JS"
```

**mini_racer-0.17.0.pre6**
```ruby
demo.rb:3:in 'block in <main>': raised in ruby (RuntimeError)
	from bla.rb:4:in 'MiniRacer::Context#eval'
	from bla.rb:4:in '<main>'
```

Is this not supported anymore?

We found this issue https://github.com/rubyjs/mini_racer/issues/44 from 2016, which discusses this feature and then got closed.
I tried versions down to 0.5.0 where it works. In versions before this i get segfaults or have issues installing.

Thanks for your help and have a nice day.

SamSaffron:
tough one @bnoordhuis not sure how we would swing this with the new serialization protocol?

bnoordhuis:
@pascalbetz I can't test it tonight but can you check if this patch produces the behavior you're after?
```diff
diff --git a/ext/mini_racer_extension/mini_racer_v8.cc b/ext/mini_racer_extension/mini_racer_v8.cc
index 38d14ce..3d21796 100644
--- a/ext/mini_racer_extension/mini_racer_v8.cc
+++ b/ext/mini_racer_extension/mini_racer_v8.cc
@@ -388,8 +388,11 @@ void v8_api_callback(const v8::FunctionCallbackInfo<v8::Value>& info)
         v8_roundtrip(st.ruby_context, &p, &n);
         if (*p == 'c') // callback reply
             break;
-        if (*p == 'e') // ruby exception pending
-            return st.isolate->TerminateExecution();
+        if (*p == 'e') { // ruby exception pending
+            auto message = v8::String::NewFromUtf8Literal(st.isolate, "Ruby exception");
[...]

PR #373, opened 2025-10-04
JS code can now catch Ruby exceptions

bnoordhuis:
Before this commit exceptions from Ruby callbacks were translated to termination exceptions that cannot be caught by JS code. Turn them into regular exceptions that can be caught.

Fixes: https://github.com/rubyjs/mini_racer/issues/357

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.c
+++ b/ext/mini_racer_extension/mini_racer_extension.c
@@ -482,12 +482,10 @@ static void des_object_ref(void *arg, uint32_t id)
 
 static void des_error_begin(void *arg)
 {
-    push(arg, rb_class_new_instance(0, NULL, rb_eRuntimeError));
 }
 
 static void des_error_end(void *arg)
 {
-    pop(arg);
 }
 
 static int collect(VALUE k, VALUE v, VALUE a)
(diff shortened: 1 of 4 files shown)


tisba:
I think the truffelruby tests are not supposed to fail 🤔

bnoordhuis:
I disabled the test for truffleruby. Ideally truffleruby and mruby should behave the same but I don't know enough about the former to say whether that's feasible.

bnoordhuis:
The error went away after a rerun. I ran valgrind locally a couple of times but I've not been able to definitely pinpoint any problems in mini_racer itself.

As I mentioned earlier, there are plenty of warnings, and mini_racer.so shows up in some of the stack traces, but the top-most frame is always some internal ruby function operating on a VALUE. I've not been able to trace any of those back to mini_racer not initializing a VALUE properly.

I can't conclusively say it's not a mini_racer issue (the data is too tenuous) but there are no obvious smoking guns.

tisba:
IIRC, @eregon took a look from time to time to mini_racer's truffleruby support, so I'm just pinging him for FYI.

bnoordhuis:
Thanks @eregon, works great. Anyone wants to LGTM this?

PR #380, opened 2025-11-25
CI: Adds Ruby 4.0.0-preview2

tisba:
Initially I wanted to add Ruby head/dev, but that's not that easy for the linux jobs as they use https://hub.docker.com/_/ruby which don't provide nightly builds.

I think it still makes sense to explicitly test Ruby 4 previews (currently 4.0.0-preview2).


tisba:
I already set a reminder for myself. At least it looks like nothing serious broke for us with Ruby 4, but let's see 🤞

PR #397, opened 2026-03-07
Suppress -Wincompatible-pointer-types when needed

knu:
LLVM Clang 22+ treats -Wincompatible-pointer-types as an error by default, which breaks compilation of mini_racer_extension.c due to the unsigned long vs uint64_t mismatch in the bigint serialization code.

Add -Wno-incompatible-pointer-types to CFLAGS when the compiler rejects an unsigned long to uint64_t pointer conversion, as a workaround until the type mismatch is properly fixed in the extension source.

Ref: https://github.com/rubyjs/mini_racer/issues/359
Ref: https://github.com/rubyjs/mini_racer/pull/361

### DIFF
--- a/ext/mini_racer_extension/extconf.rb
+++ b/ext/mini_racer_extension/extconf.rb
@@ -32,6 +32,17 @@ $CXXFLAGS += " -fms-extensions"
 
 $CXXFLAGS += " -Wno-reserved-user-defined-literal" if IS_DARWIN
 
+# LLVM Clang 22+ treats -Wincompatible-pointer-types as an error by default;
+# suppress it until the `unsigned long` vs `uint64_t` mismatch in the extension
+# is properly fixed.
+unless try_compile(<<~'C', '-Wincompatible-pointer-types')
+  #include <stdint.h>
+  void f(const uint64_t *p) { (void)p; }
+  void g(void) { unsigned long a[1]; f(a); }
+C
+  $CFLAGS += " -Wno-incompatible-pointer-types"
+end
+
 if IS_DARWIN
   $LDFLAGS.insert(0, " -stdlib=libc++ ")
 else


tisba:
Hey! I added macOS 26 (both on arm and x86) to the CI matrix: https://github.com/rubyjs/mini_racer/pull/398. It seems to compile just fine, even without your change… 🤔 The images use clang 17, according to https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md.

Maybe there is another cause for the compilation issues?

EDIT: The referenced issues mention this problem also happening on older macOS versions, not only macOS 26.

knu:
@tisba Sorry, the build process picked Clang 22 installed via Homebrew, which was the cause.  I've updated the fix.

https://github.com/llvm/llvm-project/blob/llvmorg-22.1.0/clang/docs/ReleaseNotes.rst#c-c-language-potentially-breaking-changes

tisba:
Thank you. This greatly exceeds my level of expertise though 😅 

Maybe @SamSaffron @bnoordhuis can take a look?

knu:
The added conditional basically checks if the compiler accepts `-Wincompatible-pointer-types` and if such a warning counts as a hard error, and we add `-Wno-incompatible-pointer-types` to CFLAGS if that's the case.

bnoordhuis:
Wouldn't this also solve the issue?
```diff
diff --git a/ext/mini_racer_extension/serde.c b/ext/mini_racer_extension/serde.c
index 7d5d51f..3a3d8eb 100644
--- a/ext/mini_racer_extension/serde.c
+++ b/ext/mini_racer_extension/serde.c
@@ -244,7 +244,7 @@ static void ser_num(Ser *s, double v)
 }
 
 // ser_bigint: |n| is in bytes, not quadwords
-static void ser_bigint(Ser *s, const uint64_t *p, size_t n, int sign)
+static void ser_bigint(Ser *s, const unsigned long *p, size_t n, int sign)
 {
     if (*s->err)
         return;
```

knu:
@bnoordhuis It doesn't.

```
...
compiling mini_racer_v8.cc
clang++: warning: argument unused during compilation: '-rdynamic' [-Wunused-command-line-argument]
In file included from mini_racer_extension.c:37:
./serde.c:101:27: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
  101 |     n = next_power_of_two(n);
      |         ~~~~~~~~~~~~~~~~~ ^
./serde.c:111:14: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Wshorten-64-to-32]
  111 |     b->cap = n;
      |            ~ ^
./serde.c:283:23: error: incompatible pointer types passing 'uint64_t *' (aka 'unsigned long long *') to parameter of type 'const unsigned long *' [-Wincompatible-pointer-types]
  283 |         ser_bigint(s, &t, sizeof(t), sign);
      |                       ^~
./serde.c:247:53: note: passing argument to parameter 'p' here
  247 | static void ser_bigint(Ser *s, const unsigned long *p, size_t n, int sign)
      |                                                     ^
...
```

```console
% clang++ --version
Homebrew clang version 22.1.2
Target: arm64-apple-darwin25.4.0
Thread model: posix
InstalledDir: /opt/homebrew/Cellar/llvm/22.1.2/bin
Configuration file: /opt/homebrew/Cellar/llvm/22.1.2/etc/clang/arm64-apple-darwin25.cfg
```

SamSaffron:
https://github.com/rubyjs/mini_racer/pull/434

Does this work? I think it resolves it without hacking compile - cc @bnoordhuis

SamSaffron:
merged my pr I think we compile fine now

knu:
Has a new version to address this already released?

tisba:
> Has a new version to address this already released?

as far as I can tell, not yet. Might be an oversight /cc @SamSaffron

SamSaffron:
yes I think so. it should released.

tisba:
https://github.com/rubyjs/mini_racer/commit/554923f29985418da554baf74411fa017af9b132 is not in the v0.22.0 tag ( https://github.com/rubyjs/mini_racer/compare/v0.22.0...main)

SamSaffron:
oh ... let me do a release.

knu:
Thank you for releasing 0.22.1!

PR #406, opened 2026-04-16
FEATURE: add Ruby-to-JS Uint8Array support

SamSaffron:
Introduce MiniRacer::Binary for attached Ruby callbacks that need to
return raw bytes to JavaScript. The C extension now serializes this
wrapper as a Uint8Array, and TruffleRuby is updated to mirror the same
conversion path.

Also add documentation, tests, and bump the version to 0.20.1.


davidtaylorhq (approved):
Looks good to me. Perhaps this should be `0.21.0`, since it introduces a new feature?

SamSaffron:
sure I can do a bump...

PR #410, opened 2026-05-01
Fix TruffleRuby job in CI

### LINKED ISSUE #409: TruffleRuby hang in CI
TruffleRuby is hanging in CI, I'm investigating.

_Originally posted by @tisba in https://github.com/rubyjs/mini_racer/issues/408#issuecomment-4343953728_

eregon:
Fixes https://github.com/rubyjs/mini_racer/issues/409.

https://github.com/rubyjs/mini_racer/pull/406 broke the CI job but that was unclear because the CI job hanged.
The hang is unfortunate and turns out to be a bug in Truffle (https://github.com/truffleruby/truffleruby/issues/4233).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,18 +14,14 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - "macos-14" # arm64
-          - "macos-15" # arm64
-          - "macos-26" # arm64
-          - "ubuntu-24.04"
+          - "macos-latest"
+          - "ubuntu-latest"
         ruby:
           - "truffleruby+graalvm"
 
     name: ${{ matrix.os }} - ${{ matrix.ruby }}
     runs-on: ${{ matrix.os }}
-
-    env:
-      TRUFFLERUBYOPT: "--jvm --polyglot"
+    timeout-minutes: 10
 
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v4
@@ -34,7 +30,6 @@ jobs:
       - uses: ruby/setup-ruby@4c56a21280b36d862b5fc31348f463d60bdc55d5 # v1
         with:
           ruby-version: ${{ matrix.ruby }}
-          bundler: latest # to get this fix: https://github.com/rubygems/rubygems/issues/6165
           bundler-cache: true
       - name: Install GraalVM JS component
         run: truffleruby-polyglot-get js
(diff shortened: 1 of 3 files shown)


test/mini_racer_test.rb
```diff
@@ -1226,12 +1226,12 @@ def test_large_integer
 
   def test_binary_returns_uint8array
     context = MiniRacer::Context.new
-    context.attach("add_one", ->(data) {
-      MiniRacer::Binary.new(data.bytes.map { _1 + 1 }.pack("C*"))
```
eregon, line 1230: @SamSaffron `data.bytes` is what actually caused the failure, see https://github.com/rubyjs/mini_racer/issues/409#issuecomment-4359088927.

I'm not sure what is better for converting a `Uint8Array` from JS to Ruby, should it be a Ruby String, a Ruby Array? (it could even be a foreign array on TruffleRuby which means no copying but can't do that on CRuby)

So I changed the test to not test that since the main purpose seemed rather to create `Uint8Array` from Ruby using `MiniRacer::Binary.new`.

.github/workflows/ci.yml
```diff
@@ -14,18 +14,14 @@ jobs:
...
+          - "ubuntu-latest"
         ruby:
           - "truffleruby+graalvm"
 
     name: ${{ matrix.os }} - ${{ matrix.ruby }}
     runs-on: ${{ matrix.os }}
-
-    env:
-      TRUFFLERUBYOPT: "--jvm --polyglot"
+    timeout-minutes: 10
```
tisba, line 24: That's a good idea. We should probably add those to all jobs? 🤔

eregon: We could, though it's only useful when a job hangs

test/mini_racer_test.rb
SamSaffron: trouble with ruby array is that it is inefficient, no :( the only way of getting a clean uint array in ruby that actually uses the right amount of storage is string, is there some other primitive I am missing? I don't mind the change the tests read fine.

SamSaffron: May IO::Buffer ? it is experimental which is annoying...

eregon: Right, between Ruby Array and Ruby String, String is more compact for this.
(on TruffleRuby it'll be 4 bytes per element vs 8 bytes but that's still 4x more than 1 byte per element)

A "foreign array" would use the least amount of memory and avoid copying (just a small wrapper), but I think the current version of mini_racer can't do that anymore, it serializes everything.

I'll add a test explicitly for that and implement it on TruffleRuby then.

eregon: Done

Issue #411, opened 2026-05-17
Feature request: expose V8 code cache (ScriptCompiler::CachedData) on Context#eval


ursm:
## Motivation

For applications that re-execute the same multi-MB JS bundle on every request
(e.g. SSR / per-visit isolation patterns), `Script::Compile` parsing dominates
runtime — large bundles add hundreds of ms of cumulative parse cost per visit
that has nothing to do with the script's actual work.

V8 has a built-in remedy in `v8::ScriptCompiler::CachedData`: compile once,
serialize the bytecode, and feed it back on subsequent compiles to skip parsing.
mini_racer currently always goes through the plain `v8::Script::Compile` path
(`mini_racer_v8.cc:585`), so this win isn't reachable from Ruby.

This mirrors the architecture of `MiniRacer::Snapshot#dump` / `Snapshot.load`
(#395) — opaque bytes serialized by V8, persisted by the caller, fed back on
the next boot — just at the per-script granularity instead of per-isolate.

## Proposed API

Three candidate shapes; happy to take direction on which the maintainers prefer.

**(A) kwarg on `eval`, blob returned alongside result**

```ruby
result, fresh_cache = ctx.eval(src, filename: 'bundle.js', cached_data: prev_cache)
```

Symmetric with `Snapshot#dump` / `.load`. Backward-incompatible return shape
unless gated on the kwarg being present.

**(B) kwarg on `eval`, blob fetched separately**

```ruby
result = ctx.eval(src, filename: 'bundle.js', cached_data: prev_cache)
fresh_cache = ctx.last_compiled_cache  # nil if cache was consumed
```

Fully backward-compatible signature.

**(C) explicit compile step (closer to V8 native shape)**

```ruby
script = ctx.compile(src, filename: 'bundle.js', cached_data: prev_cache)
script.cached_data  # bytes for persistence
script.run          # or implicit run via eval(script)
```

More surface area, but composable; lets callers warm the cache without running.

## Implementation sketch

- Switch `v8_eval` to `v8::ScriptCompiler::Compile` with `kConsumeCodeCache` when
  a blob is supplied; check `source.GetCachedData()->rejected` and silently
  fall back to a fresh compile on rejection.
- Without a blob, compile normally and call
  `ScriptCompiler::CreateCodeCache(unbound_script)` to produce the bytes.
- Extend the request / response payloads on the bridge (third optional
  element either way) — same pattern as the snapshot serialization.
[...]

Issue #412, opened 2026-05-18
Feature: Expose V8 ES Module API on Context (CompileModule / Instantiate / Evaluate)


ursm:
# Feature: Expose V8 ES Module API on Context (CompileModule / Instantiate / Evaluate)

## Summary

`Context#eval` compiles via `v8::Script::Compile` (classic script), so `import`/`export`
syntax is rejected. Embedders that need to evaluate ES module code must currently
work around this by lexically rewriting `import`/`export` into runtime function calls
and synthesizing live bindings via `Proxy`. **The Proxy approximation diverges from
ES module semantics in ways that cannot be patched at the JS layer** — `??`, `typeof`,
`Object.getPrototypeOf`, `Symbol.toPrimitive`, and frozen-prototype detection all
observe the Proxy rather than the underlying binding.

This issue proposes exposing V8's native module loading API
(`ScriptCompiler::CompileModule`, `Module::InstantiateModule`, `Module::Evaluate`,
the resolution callback, and the dynamic-import host callback) so embedders can
route ESM through V8 directly.

## Motivation: the workaround space is closed

[capybara-simulated](https://github.com/ursm/capybara-simulated) is an in-process
headless browser for Rails system tests built on mini_racer. It lexically rewrites
`import { x } from './foo.js'` to `const x = __csim_liveImport(mod, 'x')`, where
`__csim_liveImport` returns a function-target `Proxy` that re-reads `mod[x]` on each
access. The Proxy form is required to support circular imports and the lazy-init
pattern Rolldown's CJS-interop emits (see Bug 3 below).

Two production-blocking bugs from running Mastodon under this scheme:

### Bug 1: `??` + `typeof` lazy-init (Mastodon → react-dom `useState`)

Mastodon's `initial_state-*.js` exports settings that may be `undefined`:

```js
const autoPlayGif = __csim_liveImport(initialStateModule, 'autoPlayGif');
useState(autoPlayGif ?? false);
```

Native ESM: `autoPlayGif` is `undefined`, `?? false` short-circuits, `useState`
stores `false`. With the Proxy, `autoPlayGif ?? false` returns the Proxy (Proxies
aren't nullish), and react-dom's `typeof initial === 'function' && (initial = initial())`
lazy-init branch fires because `typeof Proxy === 'function'` — throwing `TypeError`
from the `apply` trap.

### Bug 2: `Immutable.fromJS` plain-object detection

```js
function isPlainObj(v) {
[...]

PR #413, opened 2026-05-18
FEATURE: expose V8 ScriptCompiler::CachedData via Context#compile

ursm:
## Summary

Implements #411 — exposes V8's `ScriptCompiler::CachedData` so callers can persist per-script bytecode cache and skip re-parsing large bundles on subsequent processes.

```ruby
# First process: produce the cache
ctx = MiniRacer::Context.new
script = ctx.compile(File.read("bundle.js"), filename: "bundle.js", produce_cache: true)
File.binwrite("bundle.js.cache", script.cached_data) if script.cached_data
script.run

# Later process: restore from blob, skip the parse step
cached = File.binread("bundle.js.cache")
ctx = MiniRacer::Context.new
script = ctx.compile(File.read("bundle.js"), filename: "bundle.js", cached_data: cached)
script.run
```

### API surface

- `MiniRacer::Context#compile(source, filename:, cached_data:, produce_cache:)` → `MiniRacer::Script`
- `Script#run` — executes the compiled script; safe to call multiple times
- `Script#cached_data` — bytecode blob (nil when the supplied `cached_data:` was accepted; populated on initial compile or after rejection, only when `produce_cache: true` was set)
- `Script#cache_rejected?` — boolean for cache-key invalidation telemetry
- `Script#dispose` / `Script#disposed?` — eager handle release
- `MiniRacer::V8_CACHED_DATA_VERSION_TAG` — module-level constant (populated on first `Context.new`) wrapping `v8::ScriptCompiler::CachedDataVersionTag()`; mix into cache keys so a libv8-node bump invalidates blobs automatically

## Safety constraints

Found while wiring this into a real embedder (capybara-simulated driving Discourse). All three are documented in the README; the first two are also enforced at runtime.

1. **`produce_cache:` defaults to `false`.** Passing `true` from inside a host-fn callback raises `MiniRacer::RuntimeError`. V8's `CreateCodeCache` walks live isolate state and corrupts the parser when re-entered from a JS → Ruby → JS frame; standalone repro at https://github.com/rubyjs/mini_racer/issues — TODO. Warm the cache from the top level instead.
2. **Cross-process reuse requires byte-identical snapshot data on both sides.** `MiniRacer::Snapshot.new(src).dump` is non-deterministic across processes, so feeding the same source string to two `Snapshot.new` calls produces different blobs and V8 rejects every cached_data crossing that boundary. Use `Snapshot#dump` → persist → `Snapshot.load(bytes)` instead.
3. **Cross-process reuse is incompatible with `Platform.set_flags!(:single_threaded)`.** V8's single-threaded mode embeds process-local state in the cache blob, so cached_data is always rejected when consumed in a fresh process. Same-process reuse (e.g. a Context pool) still works. Embedders that need both will need to disable `:single_threaded` for the cache-producing / cache-consuming path.

## Design notes

**Context dispose ordering**: `State::~State()` walks `st.scripts` and resets each `v8::Persistent<v8::Script>` under the existing Locker/Isolate::Scope before `isolate->Dispose()`. Handle table is owned per-State.

**Concurrency**: compile/run/dispose RPCs go through the existing `rendezvous` mutex path; the handle table is only touched from the V8 thread. The new `State::in_callback` counter is incremented in `v8_api_callback` (also V8-thread-only) and read inside `v8_compile`; no cross-thread access, no atomic needed.

**GC finalizer trade-off**: `script_free` does NOT send a dispose RPC — taking `rr_mtx` from a Ruby finalizer thread risks deadlock. Handles freed via finalizer rely on `State::~State()` walking the table at isolate teardown. Long-lived Contexts with many short-lived Scripts will accumulate handles until `Context#dispose`. Documented in README; `Script#dispose` is available for eager release.

**CachedData buffer policy**: input blob uses `BufferNotOwned` pointing at the ValueDeserializer's ArrayBuffer backing store (valid for the v8_compile call), avoiding a copy of potentially MB-sized blobs.

**Packet protocol**: new tags `'K'` (compile), `'R'` (run), `'D'` (dispose) added to `dispatch1`. `'C'` was already taken by call, hence `'K'` for compile.

Refs #411.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,14 @@
 - (Unreleased)
   - Add `Context#perform_microtask_checkpoint` to synchronously drain the V8 microtask queue, useful for spec-compliant `dispatchEvent` sequencing inside Ruby callbacks
+  - Expose V8 ScriptCompiler::CachedData via Context#compile / MiniRacer::Script (#411)
+    - script = ctx.compile(src, filename:, cached_data:, produce_cache:) → Script handle
+    - script.run replays compiled bytecode without re-parsing
+    - script.cached_data persists V8's per-script bytecode cache when produce_cache: true
+    - script.cache_rejected? reports source/version mismatches
+    - MiniRacer::V8_CACHED_DATA_VERSION_TAG for cache-key invalidation
+    - produce_cache defaults to false; passing true from inside a host-fn callback raises (V8's CreateCodeCache corrupts parser state when re-entered, so warm the cache from the top level)
+    - Cross-process reuse requires both processes to load byte-identical snapshot data via Snapshot#dump / Snapshot.load (Snapshot.new(src) is non-deterministic) and is incompatible with Platform.set_flags!(:single_threaded) (V8 embeds per-process state in the blob under single-threaded mode)
+    - TruffleRuby shim falls back to source replay (no equivalent in GraalJS)
 
 - 0.21.1 - 25-05-2026
   - Run `:single_threaded` V8 dispatches on a reusable mini_racer-owned native thread so V8 does not execute on Ruby-owned threads
--- a/README.md
+++ b/README.md
@@ -129,6 +129,91 @@ context.eval("bar()", filename: "a/bar.js")
 # …
 ```
 
+### Bytecode cache for repeated script evaluation
+
+`Context#compile` returns a `MiniRacer::Script` handle you can run multiple times,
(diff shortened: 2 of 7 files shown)


ursm:
For sequencing context: #412 (Module API) is the next planned PR but I'm holding it back until this one lands. The two share a lot of C++ surface (handle table, packet protocol, dispose ordering) so iterating patterns here once will be cheaper than rebasing #412 twice. Flagging in case it helps frame the review.

Issue #415, opened 2026-05-19
SEGV in v8::internal::wasm::StackMemory::jslimit() during exception unwind (NULL active_continuation)


ursm:
### Summary

`ctx.call` reliably SEGVs at `v8::internal::wasm::StackMemory::jslimit()` (NULL deref at `0x68`) when running a large Ember + webpack app (Discourse 3.6) that does **not** use WebAssembly. The crash happens inside V8's exception unwinder.

The same workload runs fine under `quickjs.rb`, so the trigger is V8-specific.

### Versions

- `mini_racer` 0.21.0
- `libv8-node` 24.12.0.1
- Ruby 3.4.6
- Linux x86_64 (Gentoo, glibc 2.x)

### Crash (gdb, Thread 1, signal SIGSEGV)

```
#0  0x...43f0 v8::internal::wasm::StackMemory::jslimit() const
#1  0x...2470 v8::internal::Isolate::UnwindAndFindHandler()
#2  0x...48dc v8::internal::Runtime_UnwindAndFindExceptionHandler(int, unsigned long*, v8::internal::Isolate*)
#3  0x...067d v8_Default_embedded_blob_code_
…
#9  0x...f901 v8::internal::Object::TransitionAndWriteDataProperty(
                v8::internal::LookupIterator*,
                v8::internal::DirectHandle<v8::internal::Object>,
                v8::internal::PropertyAttributes,
                v8::Maybe<v8::internal::ShouldThrow>,
                v8::internal::StoreOrigin)
```

Ruby crash report:

```
[BUG] Segmentation fault at 0x0000000000000068
Control frame:
  CFUNC :call                                                (ctx.call)
  METHOD .../v8_runtime.rb:138 in #call
  METHOD .../browser.rb:472   in #check_stale                (calls __csimAlive)
  METHOD .../node.rb:159      in #visible?
  …
  Capybara::Node::Element#visible? polling loop
```

### Analysis

`Isolate::UnwindAndFindHandler` calls `wasm::StackMemory::jslimit()` while walking up the frame stack to find the next catch handler. `jslimit()` reads `this->jslimit_` at offset `0x68`; the `this` pointer is NULL because the Isolate's `active_continuation` slot has never been initialised — the workload never executed wasm.

Trigger conditions in this report:

- Large JS bundle loaded (Discourse's main webpack chunk is ~5.7 MB).
- A JS-side exception is thrown during a property write (`Object::TransitionAndWriteDataProperty`).
[...]

PR #416, opened 2026-05-21
Run single-threaded V8 dispatches on reusable thread

sam-saffron-jarvis:
## Summary

Fixes `:single_threaded` so V8 still uses the single-threaded platform, but MiniRacer no longer runs V8 dispatches directly on the Ruby caller thread.

Instead, each context lazily starts a reusable single-threaded V8 runner thread. The important bit: the runner waits outside V8, and only enters V8 for an individual dispatch:

```text
Ruby caller thread
  rb_nogvl
    ensure reusable runner thread exists for this pid
    signal request
    wait for response / callback

runner thread
  wait outside V8
  on request:
    enter V8 with Locker/IsolateScope/HandleScope
    dispatch request
    exit V8 completely
  wait outside V8 again
```

JS→Ruby callbacks rendezvous back to the original Ruby caller thread through the existing condition-variable protocol, so the V8 runner thread does not directly reacquire the GVL or run Ruby code.

This addresses rubyjs/mini_racer#415. In that workload, the previous same-Ruby-thread `:single_threaded` path could crash during V8 exception unwinding in `v8::internal::wasm::StackMemory::jslimit()`. The crash appears to depend on the embedding shape where V8 runs on a Ruby thread and callbacks re-enter Ruby/GVL from inside the V8 stack.

No extra public option is added. `MiniRacer::Platform.set_flags!(:single_threaded)` keeps the existing public API.

## Fork behavior

This preserves the important historical property of `:single_threaded`: an inherited context can still be used after `fork`.

The reusable runner is tracked per process id. If a context is inherited into a child process, the parent runner thread is gone, so the next dispatch notices the pid change and starts a new runner thread in the child.

The runner does not hold V8's `Locker` or `Isolate::Scope` while idle, so the normal idle-at-fork case does not inherit a V8 thread parked inside the isolate.

## Changes

- Replace direct same-Ruby-thread `v8_single_threaded_enter(...)` dispatches with a reusable per-context runner thread.
- The runner enters/exits V8 around each dispatch instead of holding V8 state while idle.
- Make `v8_roundtrip()` use the condition-variable rendezvous path for `:single_threaded` too.
- Keep Ruby callbacks running on the original Ruby caller thread, not on the V8 runner thread.
- Track the runner pid and recreate the runner after fork.
- Keep the existing single-threaded persistent V8 handles, so contexts remain alive between calls and across ordinary fork boundaries.

## Validation

MiniRacer test suite:

```text
103 runs, 188 assertions, 0 failures, 0 errors, 3 skips
```

Forking smoke test:

```sh
bundle exec ruby test/test_forking.rb
# exit 0
```

Fork stress:

```sh
for i in $(seq 1 10); do bundle exec ruby test/test_forking.rb || exit 1; done
# fork-stress-ok
```

Single-threaded Ruby callback smoke:

```sh
bundle exec ruby -Ilib -e 'require "mini_racer"; MiniRacer::Platform.set_flags! :single_threaded; c=MiniRacer::Context.new; c.attach("rb", proc { |x| x + 1 }); raise unless c.eval("rb(41)") == 42; puts "callback ok"'
# callback ok
```

Standalone rubyjs/mini_racer#415 replay now completes with the normal existing flag:

```ruby
MiniRacer::Platform.set_flags!(:single_threaded)
```

```text
visited, current_url=http://www.example.com/about status=200
checking .about__stats-item.site-creation-date span text="Created 2 months ago" iterations=1
iteration 0: found element text="Created 2 months ago"; asking has_text?/visible?
done without crash
```

The original Discourse repro from #415 also passes unchanged:

```text
1 example, 0 failures
```

### DIFF
--- a/ext/mini_racer_extension/mini_racer_extension.c
+++ b/ext/mini_racer_extension/mini_racer_extension.c
@@ -4,6 +4,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <pthread.h>
+#include <unistd.h>
 #include <math.h>
 
 #if defined(__linux__) && !defined(__GLIBC__)
@@ -136,6 +137,9 @@ typedef struct Context
     VALUE exception;   // pending exception or Qnil
     Buf req, res;      // ruby->v8 request/response, mediated by |mtx| and |cv|
     Buf snapshot;
+    pthread_t single_threaded_thr;
+    pid_t single_threaded_pid;
+    int single_threaded_thr_started;
     // |rr_mtx| stands for "recursive ruby mutex"; it's used to exclude
     // other ruby threads but allow reentrancy from the same ruby thread
     // (think ruby->js->ruby->js calls)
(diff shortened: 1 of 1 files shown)


SamSaffron:
directly merged this.

PR #418, opened 2026-05-27
FEATURE: add Context#perform_microtask_checkpoint

ursm:
## Summary

- Adds `MiniRacer::Context#perform_microtask_checkpoint`, a thin wrapper around V8's `MicrotasksScope::PerformCheckpoint(isolate)`.
- Lets Ruby callbacks invoked during `context.call` / `context.eval` synchronously drain the microtask queue mid-execution.

## Motivation

V8's default `kAuto` microtask policy drains the queue when JS script execution returns to the embedder. When JS calls a host (Ruby) function that itself calls back into JS — for example, a synthetic `dispatchEvent` simulating browser-style event propagation — all the JS runs inside one script execution, so microtasks queued by listener N do not drain before listener N+1. This diverges from the HTML "clean up after running script" spec, where the UA's dispatcher runs each listener via a fresh microtask checkpoint.

Currently the only escape hatch is `pump_message_loop`, which drags in `PumpMessageLoop` / `RunIdleTasks` and is bool-returning. A dedicated primitive is cleaner.

## API

```ruby
context.attach('drain', -> { context.perform_microtask_checkpoint })
context.eval(<<~JS)
  Promise.resolve().then(() => log('microtask-fired'));
  log('before-drain');
  drain();
  log('after-drain');
JS
# Order: before-drain, microtask-fired, after-drain
```

## Notes

- The implementation mirrors `pump_message_loop`'s structure (HandleScope + verbose TryCatch). It deliberately does **not** call `CancelTerminateExecution` — if a microtask trips OOM via the GC epilogue or watchdog termination fires, the enclosing `v8_call` / `v8_eval` frame surfaces the error to Ruby as usual.
- Works in both threaded and `:single_threaded` modes (uses the same dispatch tag mechanism).
- Tested at top level and from inside a callback during `context.eval`.

## Test plan

- [x] `test_perform_microtask_checkpoint_returns_nil`
- [x] `test_perform_microtask_checkpoint_drains_from_callback`
- [x] Full existing test suite passes (113 runs, 0 failures)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+- (Unreleased)
+  - Add `Context#perform_microtask_checkpoint` to synchronously drain the V8 microtask queue, useful for spec-compliant `dispatchEvent` sequencing inside Ruby callbacks
+
 - 0.21.1 - 25-05-2026
   - Run `:single_threaded` V8 dispatches on a reusable mini_racer-owned native thread so V8 does not execute on Ruby-owned threads
   - Stop and join the reusable `:single_threaded` runner when contexts are disposed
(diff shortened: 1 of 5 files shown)


SamSaffron (approved):

SamSaffron:
nice change! 

we need to start thinking about docs, the readme is getting very big but we certainly want some place where everything is properly documented. can you do a follow up pr to update the readme?

PR #419, opened 2026-05-28
DOC: document Context#perform_microtask_checkpoint

ursm:
Follow-up to #418 as requested.

Adds a **Microtask checkpoints** subsection to the README:

- Explains V8's default auto-drain (`Promise.resolve().then(...)` "just works" for simple scripts).
- Documents `Context#perform_microtask_checkpoint` and when you actually need it — the synchronous-Ruby-callback case (e.g. chained `dispatchEvent` listeners running in one `context.call`).
- Both code samples were run through `mini_racer` locally; outputs in the comments are real.

Placed after **Function call**, before **Performance**, matching the existing flat h3-section pattern in that area. Kept the section short given your note about README size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,38 @@ Performance is slightly better than running `context.eval("hello('George')")` si
 * compilation of eval'd string is avoided
 * function arguments don't need to be converted to JSON
 
+### Microtask checkpoints
+
+V8 drains its microtask queue (e.g. callbacks queued via `Promise.resolve().then(...)`) automatically when script execution returns to the embedder, so most code "just works":
+
+```ruby
+context = MiniRacer::Context.new
+context.eval(<<~JS)
+  let x = 0;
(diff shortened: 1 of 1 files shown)


SamSaffron:
CI failure is strange, it just skipped 3 specs ... not exit 1 worthy.

tisba:
@SamSaffron in https://github.com/rubyjs/mini_racer/pull/417 I noticed that Ruby 3.1 consistently seems to hang in CI. The other failures are more intermittent and might also be some Github issues? 🤷

SamSaffron:
Ruby 3.2 is EOL, we should drop 3.2/3.1 from CI and just test with 4/3.4/3.3

tisba:
> Ruby 3.2 is EOL, we should drop 3.2/3.1 from CI and just test with 4/3.4/3.3

https://github.com/rubyjs/mini_racer/pull/425

PR #425, opened 2026-06-01
Bump Ruby version to >=3.3

tisba:
As per our policy, mini_racer only supports non-EOL'd versions of Ruby. Currently Ruby 3.3+ is supported, see https://www.ruby-lang.org/en/downloads/branches/.


tisba:
Not sure why truffleruby is failing now in CI, this PR doesn't touch anything in that regard.

/cc jFYI @eregon

SamSaffron:
looks good!

PR #433, opened 2026-06-24
Feature: Faster handling of int arrays

SamSaffron:
New numbers post PR are: 

Default platform is now faster than v0.21.0 for these rows:

callback/1000_echo_int  v0.21.0 3.278 ms  working 2.449 ms
callback/1000_add_ints  v0.21.0 3.845 ms  working 2.567 ms

not sure if it is worth the complexity or not so will sit on it


SamSaffron:
not sure complexity is worth it yet.

